PR #3, opened 2019-02-19
Updated the homepage to include 'Hello from Team 9'.

michaelsafyan:


src/main/webapp/index.html
```diff
@@ -33,7 +33,7 @@
         <h1>CodeU 2019 Starter Project</h1>
         <div>
             <p>
-                Welcome to Code U!
+                Welcome to Code U! Hello from Team 9!
```
markser, line 36: This is good, but maybe you want to add a real team name instead of Team 9!

markser (approved):
Looks good, but we should follow up with the correct team name.

PR #4, opened 2019-02-20
Completed steps 1-4 for Direct Messages. Implemented recipient field.

shamailah:
Added a recipient field for the Messages object.

### DIFF
--- a/src/main/java/com/google/codeu/data/Datastore.java
+++ b/src/main/java/com/google/codeu/data/Datastore.java
@@ -42,6 +42,7 @@ public class Datastore {
     messageEntity.setProperty("user", message.getUser());
     messageEntity.setProperty("text", message.getText());
     messageEntity.setProperty("timestamp", message.getTimestamp());
+    messageEntity.setProperty("recipient", message.getRecipient());
 
     datastore.put(messageEntity);
   }
@@ -67,8 +68,9 @@ public class Datastore {
         UUID id = UUID.fromString(idString);
         String text = (String) entity.getProperty("text");
         long timestamp = (long) entity.getProperty("timestamp");
+        String recipient = (String) entity.getProperty("recipient");
+        Message message = new Message(id, user, text, timestamp, recipient);
 
-        Message message = new Message(id, user, text, timestamp);
         messages.add(message);
       } catch (Exception e) {
         System.err.println("Error reading message.");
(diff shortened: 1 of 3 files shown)


markser (approved):
Everything is good here!
Might want to consider changing variable name to something like recipientMessage to clarify even more.

michaelsafyan (approved):
I have no objections to the technical change, itself, but do have some high-level feedback...

It is a little unclear to me which fields in `Message` are required and which ones are optional. Are you assuming that `recipient` is always set and non-null (all messages are direct to some other user) or that it is optional (and, if null, that the message is broadcast publicly)? You might want to consider clarifying the assumptions in documentation and in code (such as by using the `@Nullable` or `Optional<>` to clarify nullability/optionality, adding precondition-checking where appropriate, etc.).

It's a good idea to add unit tests for code changes, although the burden is a little higher now given that there aren't existing examples on which to draw and add. I don't think that adding unit tests should block this change, but I would encourage the creation of unit tests soon; the more code that piles up without unit tests, the harder it becomes to add unit tests.

michaelsafyan:
> Everything is good here!
> Might want to consider changing variable name to something like recipientMessage to clarify even more.

Feedback on this feedback for @markser ...

## General Naming Feedback

I would recommend giving naming feedback only if the name is confusing or if the name goes against common patterns, creating a stylistic inconsistency with other names. In both cases, it is a good idea to explain the basis of the confusion or the inconsistency.

## Feedback on the Specific Recommendation

It looks to me like `text` contains the body of the message (i.e. the message) and `recipient` contains the person to whom the message is to be sent. In the proposed variable name `recipientMessage`, "message" is the noun and "recipient" becomes a modifier; that makes "recipientMessage" imply that it no longer describes a person who receives the message, but the body of the message itself (and then it begs the question of how it differs from `text` and where the person receiving the message is stored). So, in my view, the proposed recommendation would actually increase confusion.

However, there is an area of ambiguity that would be reasonable to comment on and provide a naming suggestion about. That is, it is unclear what the format of the data stored in the field will be. For example, is it an email address? A phone number? A user ID? There are different ways to refer to a user.

While, personally, I think that `recipient` is short and sweet and such ambiguity ought to be resolved by documentation or by a more specific data type (e.g. an `EmailAddress` or `UserId` data type), in terms of giving naming recommendations, it would not be unreasonable to recommend `recipientEmailAddress` or `recipientUserId` to resolve that ambiguity. I would avoid naming recommendations, though, that don't go directly to a point of confusion, that don't resolve an inconsistency, or may add confusion.

PR #6, opened 2019-02-20
Creating a new servlet for the StatsPage

markser:
This is creating a new servlet for the StatsPage. This completes up to step 2.

### DIFF
--- a/src/main/java/com/google/codeu/data/Datastore.java
+++ b/src/main/java/com/google/codeu/data/Datastore.java
@@ -86,10 +86,10 @@ public class Datastore {
   }
 
   /** Returns the total number of messages for all users. */
-  public int getTotalMessageCount(){
+  public int getTotalMessageCount() {
     Query query = new Query("Message");
     PreparedQuery results = datastore.prepare(query);
-    return results.countEntities(FetchOptions.Builder.withLimit(1000));
-}
+    return results.countEntities(FetchOptions.Builder.withDefaults());
+  }
 
 }
--- a/src/main/java/com/google/codeu/data/Datastore.java
+++ b/src/main/java/com/google/codeu/data/Datastore.java
@@ -27,6 +27,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
+/** import for Fetch Options */
+import com.google.appengine.api.datastore.FetchOptions;
+
 /** Provides access to the data stored in Datastore. */
 public class Datastore {
 
@@ -81,4 +84,12 @@ public class Datastore {
 
     return messages;
   }
+
+  /** Returns the total number of messages for all users. */
+  public int getTotalMessageCount(){
+    Query query = new Query("Message");
+    PreparedQuery results = datastore.prepare(query);
+    return results.countEntities(FetchOptions.Builder.withLimit(1000));
+}
+
 }
--- a/src/main/java/com/google/codeu/servlets/StatsPageServlet.java
+++ b/src/main/java/com/google/codeu/servlets/StatsPageServlet.java
@@ -2,6 +2,8 @@ package com.google.codeu.servlets;
 
 import java.io.IOException;
 
+import javax.servlet.ServletConfig;
+import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -19,13 +21,15 @@ public class StatsPageServlet extends HttpServlet{
   private Datastore datastore;
 
   @Override
-  public void init() {
+  public void init(ServletConfig config) throws ServletException {
+    super.init(config);
     datastore = new Datastore();
   }
 
   /**
    * Responds with site statistics in JSON.
    */
+
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException {
--- a/src/main/java/com/google/codeu/servlets/StatsPageServlet.java
+++ b/src/main/java/com/google/codeu/servlets/StatsPageServlet.java
@@ -1,0 +1,40 @@
+package com.google.codeu.servlets;
+
+
+import java.io.IOException;
+
+
+import javax.servlet.annotation.WebServlet;
+
+import javax.servlet.http.HttpServlet;
+
+import javax.servlet.http.HttpServletRequest;
+
+import javax.servlet.http.HttpServletResponse;
+
+
+/**
+
+ * Responds with a hard-coded message for testing purposes.
+
+ */
+
+@WebServlet("/stats")
+
+public class StatsPageServlet extends HttpServlet{
+
+
+
+ @Override
+
+ public void doGet(HttpServletRequest request, HttpServletResponse response)
+
+   throws IOException {
+
+
+
+  response.getOutputStream().println("hello world");
+
+ }
+
+}
--- a/src/main/java/com/google/codeu/servlets/StatsPageServlet.java
+++ b/src/main/java/com/google/codeu/servlets/StatsPageServlet.java
@@ -1,40 +1,19 @@
 package com.google.codeu.servlets;
 
-
 import java.io.IOException;
 
-
-import javax.servlet.annotation.WebServlet;
-
 import javax.servlet.http.HttpServlet;
-
 import javax.servlet.http.HttpServletRequest;
-
 import javax.servlet.http.HttpServletResponse;
-
-
 /**
-
  * Responds with a hard-coded message for testing purposes.
-
  */
-
-@WebServlet("/stats")
-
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/codeu/servlets/StatsPageServlet.java
```diff
@@ -0,0 +1,40 @@
+package com.google.codeu.servlets;
+
+
+import java.io.IOException;
+
+
+import javax.servlet.annotation.WebServlet;
```
michaelsafyan: Please delete the `WebServlet` import and remove the `@WebServlet` annotation below.

michaelsafyan (changes requested):
Can you add an entry in `web.xml` for the servlet?

src/main/java/com/google/codeu/servlets/StatsPageServlet.java
markser: The changes have been made.

markser:
Removed the webservlet import and the @webservlet annotation, also made the servelt/servlet-mapping entry in the web.xml file.

Current '/stats' page is output is 'hello world'.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -81,4 +84,12 @@ public void storeMessage(Message message) {
 
     return messages;
   }
+
+  /** Returns the total number of messages for all users. */
+  public int getTotalMessageCount(){
+    Query query = new Query("Message");
+    PreparedQuery results = datastore.prepare(query);
+    return results.countEntities(FetchOptions.Builder.withLimit(1000));
```
michaelsafyan: It looks like this limit would make it so that the app would always return "1000" if it had very high usage. While it may make sense to specify the "chunkSize", you probably do not want a fixed limit so that you actually give the total count.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -81,4 +84,12 @@ public void storeMessage(Message message) {
 
     return messages;
   }
+
+  /** Returns the total number of messages for all users. */
+  public int getTotalMessageCount(){
+    Query query = new Query("Message");
```
michaelsafyan, line 90: As I understand it, calling `Query.setKeysOnly()` on the query object will ensure that the query fetches only the keys for the entities and not the underlying data, making the query more efficient.

src/main/java/com/google/codeu/servlets/StatsPageServlet.java
```diff
@@ -0,0 +1,41 @@
...
+ * 
+ * This uses the init() function that creates and store a Datastore instance,
+ * then doGet() calls the function getTotalMessageCount() in the Datastore.java
+ * and create a JSON that countains the message count
+ */
+public class StatsPageServlet extends HttpServlet{
+  private Datastore datastore;
+
+  @Override
+  public void init() {
```
michaelsafyan: Is there a reason that you do this initialization here rather than statically or via a constructor? (I'm not saying that you need to do it differently, I'd just like to better understand this particular design decision).

src/main/java/com/google/codeu/servlets/StatsPageServlet.java
```diff
@@ -0,0 +1,41 @@
...
+    datastore = new Datastore();
+  }
```
michaelsafyan, line 27: Don't forget to call `super.init();` -- you don't want to accidentally omit necessary initialization of the base class.

src/main/webapp/js/stats-loader.js
```diff
@@ -0,0 +1,23 @@
...
+    const url = '/stats';
+    fetch(url).then((response) => {
+      return response.json();
+    }).then((stats) => {
+      const statsContainer = document.getElementById('stats-container');
+      statsContainer.innerHTML = '';
+      const messageCountElement = buildStatElement('Message count: ' + stats.messageCount);
+      statsContainer.appendChild(messageCountElement);
+    });
+  }
```
michaelsafyan, line 12: The indentation that you have here is a little confusing; it makes it look like the code below is still within 'fethStats'.

src/main/webapp/js/stats-loader.js
```diff
@@ -0,0 +1,23 @@
+// Fetch stats and display them in the page.
```
michaelsafyan, line 1: I'd suggest the following formulation:

```
"use strict";
(function(){
   // ... your code here ...
})();
```

This formulation will give you a bit more checking and will also ensure that functions/variables are encapsulated and private unless you explicitly export them (e.g. by assigning them to a property of the global `window` object).

src/main/webapp/js/stats-loader.js
```diff
@@ -0,0 +1,23 @@
...
+      const statsContainer = document.getElementById('stats-container');
+      statsContainer.innerHTML = '';
+      const messageCountElement = buildStatElement('Message count: ' + stats.messageCount);
+      statsContainer.appendChild(messageCountElement);
+    });
+  }
+
+  function buildStatElement(statString){
+   const statElement = document.createElement('p');
+   statElement.appendChild(document.createTextNode(statString));
```
michaelsafyan, line 16: While this formulation is fine, it is a little simpler to assign to the `textContent` property as in:

```
statElement.textContent = statstring;
```

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

src/main/webapp/js/stats-loader.js
```diff
@@ -0,0 +1,23 @@
...
+   return statElement;
+  }
+
+  // Fetch data and populate the UI of the page.
+  function buildUI(){
```
michaelsafyan, line 21: Is there a reason that you have an alias like this? Why not just call it using the `fetchStats` function name and remove this alias?

src/main/webapp/stats.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Stats</title>
+        <link rel="stylesheet" href="/css/main.css">
+        <link rel="stylesheet" href="/css/stats.css">
+        <script src="/js/stats-loader.js"></script>
+    </head>
+    <body onload="buildUI()">
```
michaelsafyan, line 9: We can leave this as-is for the purpose of this change, but it seems to me that the approach taken to loading the JavaScript and issuing this fetch could be more efficient...

Right now, you are waiting until the `<body>` has finished loading to issue the call to `/stats`, but you could actually begin to fetch `/stats` fetch immediately and make it so that only the `.then` piece which puts the result into the page needs to wait for the page to finish loading.

michaelsafyan (approved):
Thanks. I left some feedback/suggestions. I'd recommend following up on those before submitting, but I don't want to block you from merging this pull request if you'd prefer to follow-up on some of those items in a separate, follow-on pull request. If you submit the pull request, please make sure to deploy the changes so that the changes are live on the App Engine application.

src/main/java/com/google/codeu/servlets/StatsPageServlet.java
markser: I believe the intention was doing the initialization every time that the stats servlet is created. I actually dont know why, it could be later on that it becomes pertinent when we start implementing more things that are user focused.

src/main/webapp/js/stats-loader.js
markser: For this instance, there is no reason. I was thinking that as we add more things to the stats page, that having  a general function "buildUI" is better than having the html page calling all of the different functions

src/main/java/com/google/codeu/servlets/StatsPageServlet.java
michaelsafyan: I suspect that the example code might do it this way for testing reasons, though -- if that is the case -- it would probably make sense to change this to:

```
public void init() {
   super.init();
   datastore = createDatastore();
}

/*
 * Instantiates a connection to a datastore. In tests, override this with a method that
 * returns a fake/mock datastore object.
 */
protected Datastore createDatastore() {
   return new Datastore();
}
```

As-is, this code does not provide a way to replace the `datastore` with a fake/mock version. Given that, it doesn't make a whole lot of sense to instantiate a separate `Datastore` per servlet rather than having simply:

```
   private static Datastore datastore = new Datastore();
```

That being said, I would encourage going with the former rather than the latter (because testing is important and we should write code in a manner that lends itself to being easily tested).

src/main/webapp/js/stats-loader.js
michaelsafyan: Let's follow the "YAGNI" (you ain't gonna need it) principle; I concur that it would make sense for the HTML code to call only a single function to populate the entire UI, but let's start with it just calling the one function (without an alias); if you decide that you need multiple functions, you can always introduce this layer of indirection when it becomes necessary.

PR #9, opened 2019-02-25
Added an About Me Servlet class to display something in /about

paulinagzd:
Step 2 of familiarization project week 1. Implemented a new class that displays a simple message to test.

### DIFF
--- a/.idea/jsLibraryMappings.xml
+++ b/.idea/jsLibraryMappings.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="JavaScriptLibraryMappings">
+    <includedPredefinedLibrary name="HTTP Response Handler" />
+  </component>
+</project>
--- a/src/main/java/com/google/codeu/data/Datastore.java
+++ b/src/main/java/com/google/codeu/data/Datastore.java
@@ -81,4 +81,33 @@ public class Datastore {
 
     return messages;
   }
+
+    /** Stores the User in Datastore. */
+    public void storeUser(User user) {
+        Entity userEntity = new Entity("User", user.getEmail());
+        userEntity.setProperty("email", user.getEmail());
+        userEntity.setProperty("aboutMe", user.getAboutMe());
+        datastore.put(userEntity);
+    }
+
+    /**
+     * Returns the User owned by the email address, or
+     * null if no matching User was found.
+     */
+    public User getUser(String email) {
+
+        Query query = new Query("User")
+                .setFilter(new Query.FilterPredicate("email", FilterOperator.EQUAL, email));
+        PreparedQuery results = datastore.prepare(query);
+        Entity userEntity = results.asSingleEntity();
+        if(userEntity == null) {
+            return null;
+        }
+
+        String aboutMe = (String) userEntity.getProperty("aboutMe");
+        User user = new User(email, aboutMe);
+
+        return user;
+    }
+
 }
--- a/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
+++ b/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
@@ -7,16 +7,59 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.codeu.data.Datastore;
+
 /**
- * Responds with a hard-coded message for testing purposes.
+ * Handles fetching and saving user data.
  */
 @WebServlet("/about")
-public class AboutMeServlet extends HttpServlet{
+public class AboutMeServlet extends HttpServlet {
 
+    private Datastore datastore;
+
+    @Override
+    public void init() {
+        datastore = new Datastore();
+    }
+
+    /**
+     * Responds with the "about me" section for a particular user.
+     */
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response)
             throws IOException {
 
-        response.getOutputStream().println("hello world");
+        response.setContentType("text/html");
+
+        String user = request.getParameter("user");
+
+        if(user == null || user.equals("")) {
+            // Request is invalid, return empty response
+            return;
+        }
+
+        String aboutMe = "This is " + user + "'s about me.";
+
+        response.getOutputStream().println(aboutMe);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response)
+            throws IOException {
+
+        UserService userService = UserServiceFactory.getUserService();
+        if (!userService.isUserLoggedIn()) {
+            response.sendRedirect("/index.html");
+            return;
+        }
+
+        String userEmail = userService.getCurrentUser().getEmail();
+
+        System.out.println("Saving about me for " + userEmail);
+        // TODO: save the data
+
+        response.sendRedirect("/user-page.html?user=" + userEmail);
     }
 }
--- a/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
+++ b/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
@@ -1,0 +1,22 @@
+package com.google.codeu.servlets;
+
+import java.io.IOException;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Responds with a hard-coded message for testing purposes.
+ */
+@WebServlet("/about")
+public class AboutMeServlet extends HttpServlet{
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response)
+            throws IOException {
+
+        response.getOutputStream().println("hello world");
+    }
+}
--- a/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
+++ b/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
@@ -18,7 +18,7 @@ import com.google.codeu.data.Datastore;
  * Handles fetching and saving user data.
  */
 public class AboutMeServlet extends HttpServlet {
-
+    // removed WebServlet annotation to implement the url-mapping in web.xml
     private Datastore datastore;
 
     @Override
@@ -32,7 +32,6 @@ public class AboutMeServlet extends HttpServlet {
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response)
             throws IOException {
-
         response.setContentType("text/html");
 
         String user = request.getParameter("user");
--- a/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
+++ b/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
@@ -2,6 +2,9 @@ package com.google.codeu.servlets;
 
 import java.io.IOException;
 
+/*
+import javax.servlet.annotation.WebServlet;
+ */
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -14,7 +17,6 @@ import com.google.codeu.data.Datastore;
 /**
  * Handles fetching and saving user data.
  */
-@WebServlet("/about")
 public class AboutMeServlet extends HttpServlet {
 
     private Datastore datastore;
--- a/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
+++ b/src/main/java/com/google/codeu/servlets/AboutMeServlet.java
@@ -13,6 +13,10 @@ import javax.servlet.http.HttpServletResponse;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
 import com.google.codeu.data.Datastore;
+import com.google.codeu.data.User;
+
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Whitelist;
 
 /**
  * Handles fetching and saving user data.
@@ -41,9 +45,17 @@ public class AboutMeServlet extends HttpServlet {
             return;
         }
 
-        String aboutMe = "This is " + user + "'s about me.";
+        /**
+         This shows the user's about me through our User class.
+         */
 
-        response.getOutputStream().println(aboutMe);
+        User userData = datastore.getUser(user);
+
+        if(userData == null || userData.getAboutMe() == null) {
+            return;
+        }
+
+        response.getOutputStream().println(userData.getAboutMe());
     }
 
     @Override
@@ -58,8 +70,16 @@ public class AboutMeServlet extends HttpServlet {
 
         String userEmail = userService.getCurrentUser().getEmail();
 
+        /** Fetching about me from User class (SANITIZED) */
+        String aboutMe = Jsoup.clean(request.getParameter("about-me"), Whitelist.none());
+
+        User user = new User(userEmail, aboutMe);
+        datastore.storeUser(user);
+
+        /**
         System.out.println("Saving about me for " + userEmail);
         // TODO: save the data
+        */
 
         response.sendRedirect("/user-page.html?user=" + userEmail);
     }
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -36,23 +36,33 @@
     <servlet-class>com.google.codeu.servlets.MessageServlet</servlet-class>
   </servlet>
 
+  <servlet>
+    <servlet-name>aboutme</servlet-name>
+    <servlet-class>com.google.codeu.servlets.AboutMeServlet</servlet-class>
+  </servlet>
+
   <servlet-mapping>
-    <url-pattern>/login</url-pattern>
     <servlet-name>login</servlet-name>
+    <url-pattern>/login</url-pattern>
   </servlet-mapping>
 
   <servlet-mapping>
-    <url-pattern>/login-status</url-pattern>
     <servlet-name>login-status</servlet-name>
+    <url-pattern>/login-status</url-pattern>
   </servlet-mapping>
 
   <servlet-mapping>
-    <url-pattern>/logout</url-pattern>
     <servlet-name>logout</servlet-name>
+    <url-pattern>/logout</url-pattern>
   </servlet-mapping>
 
   <servlet-mapping>
-    <url-pattern>/messages</url-pattern>
     <servlet-name>messages</servlet-name>
+    <url-pattern>/messages</url-pattern>
+  </servlet-mapping>
+
+  <servlet-mapping>
+    <servlet-name>aboutme</servlet-name>
+    <url-pattern>/about</url-pattern>
   </servlet-mapping>
 </web-app>
--- a/src/main/webapp/js/user-page-loader.js
+++ b/src/main/webapp/js/user-page-loader.js
@@ -103,7 +103,7 @@ function buildUI() {
 function fetchAboutMe(){
     const url = '/about?user=' + parameterUsername;
     fetch(url).then((response) => {
-        return response.text();
+        return response.json();
     }).then((aboutMe) => {
         const aboutMeContainer = document.getElementById('about-me-container');
         if(aboutMe == ''){
--- a/src/main/webapp/js/user-page-loader.js
+++ b/src/main/webapp/js/user-page-loader.js
@@ -44,6 +44,7 @@ function showMessageFormIfViewingSelf() {
           messageForm.classList.remove('hidden');
         }
       });
+    document.getElementById('about-me-form').classList.remove('hidden');
 }
 
 /** Fetches messages and add them to the page. */
@@ -95,4 +96,21 @@ function buildUI() {
   setPageTitle();
   showMessageFormIfViewingSelf();
   fetchMessages();
+  fetchAboutMe();
+}
+
+/** Fetches about me data */
+function fetchAboutMe(){
+    const url = '/about?user=' + parameterUsername;
+    fetch(url).then((response) => {
+        return response.text();
+    }).then((aboutMe) => {
+        const aboutMeContainer = document.getElementById('about-me-container');
+        if(aboutMe == ''){
+            aboutMe = 'This user has not entered any information yet.';
+        }
+
+        aboutMeContainer.innerHTML = aboutMe;
+
+    });
 }
--- a/src/main/webapp/js/user-page-loader.js
+++ b/src/main/webapp/js/user-page-loader.js
@@ -103,7 +103,8 @@ function buildUI() {
 function fetchAboutMe(){
     const url = '/about?user=' + parameterUsername;
     fetch(url).then((response) => {
-        return response.json();
+        /** catches an error with a .json() object, changing to .text() for testing purposes */
+        return response.text();
     }).then((aboutMe) => {
         const aboutMeContainer = document.getElementById('about-me-container');
         if(aboutMe == ''){
--- a/src/main/webapp/user-page.html
+++ b/src/main/webapp/user-page.html
@@ -42,13 +42,5 @@ limitations under the License.
 
     <div id="message-container">Loading...</div>
 
-    <div id="about-me-container">Loading...</div>
-    <div id="about-me-form" class="hidden">
-      <form action="/about" method="POST">
-        <textarea name="about-me" placeholder="about me" rows=4 required></textarea>
-        <br/>
-        <input type="submit" value="Submit">
-      </form>
-    </div>
   </body>
 </html>
--- a/src/main/webapp/user-page.html
+++ b/src/main/webapp/user-page.html
@@ -42,5 +42,13 @@ limitations under the License.
 
     <div id="message-container">Loading...</div>
 
+    <div id="about-me-container">Loading...</div>
+    <div id="about-me-form" class="hidden">
+      <form action="/about" method="POST">
+        <textarea name="about-me" placeholder="about me" rows=4 required></textarea>
+        <br/>
+        <input type="submit" value="Submit">
+      </form>
+    </div>
   </body>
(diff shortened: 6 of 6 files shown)


.idea/jsLibraryMappings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
michaelsafyan, line 1: Can you update the ".gitignore" file to cause IDE-specific files like this to be ignored and remove this file from the pull request?

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
+package com.google.codeu.servlets;
+
+import java.io.IOException;
+
+import javax.servlet.annotation.WebServlet;
```
michaelsafyan, line 8: Please remove the `WebServlet` import.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.codeu.data.Datastore;
+
+/**
+ * Handles fetching and saving user data.
+ */
+@WebServlet("/about")
```
michaelsafyan: Please remove the `WebServlet` annotation; instead, please update `web.xml` to control the URL mappings.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+public class AboutMeServlet extends HttpServlet {
+
+    private Datastore datastore;
+
+    @Override
+    public void init() {
+        datastore = new Datastore();
+    }
```
michaelsafyan, line 31: Don't forget to call "super.init();" within your override.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Responds with the "about me" section for a particular user.
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response)
+            throws IOException {
+
+        response.setContentType("text/html");
```
michaelsafyan, line 39: The output that you produce below is plain text. Either use "text/plain" (the MIME type for plain text), or update the response below to output HTML as in:

````
<!DOCTYPE html>
<html>
  <!-- ... more HTML content ... -->
</html>
````

File: src/main/java/com/google/codeu/servlets/AboutMeServlet.java [hunk 6]
```diff
@@ -0,0 +1,65 @@
...
+
+        String user = request.getParameter("user");
+
+        if(user == null || user.equals("")) {
+            // Request is invalid, return empty response
```
michaelsafyan, line 44: One should generally set an appropriate HTTP status code in error cases. For example, this probably should be a 404 Not Found or 400 Bad Request.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Responds with the "about me" section for a particular user.
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response)
+            throws IOException {
+
+        response.setContentType("text/html");
+
+        String user = request.getParameter("user");
```
michaelsafyan, line 41: Do I understand correctly that a single logged-in user is allowed to view the "about me" page of a number of different users? If so, I found this a little confusing in that I was assuming in a lot of the code below that "user" was referring to the currently logged-in user, when it is actually referring to the user whose "about me" page is being viewed. You might want to add a comment and give a name to distinguish between the two different users in question (the logged-in user vs the page subject).

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+
+        if(user == null || user.equals("")) {
+            // Request is invalid, return empty response
+            return;
+        }
+
```
michaelsafyan, line 47: Should there be a check that this user is one that actually exists in the database? What about a check that this user's "about me" profile is public and visible to the currently logged-in user?

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.codeu.data.Datastore;
+
+/**
+ * Handles fetching and saving user data.
+ */
+@WebServlet("/about")
+public class AboutMeServlet extends HttpServlet {
```
michaelsafyan, line 24: It looks like the servlet operates primarily on HTML pages. 

To allow for programmatic usage (e.g. with an Android or iOS app, with JavaScript, or with some other client-code), it is a good idea to expose the functionality using a REST-ful API (i.e. operating on JSON rather than HTML) and to render the content on the client (e.g. using JavaScript) or to have a second servlet that does HTML rendering on the server in addition to the REST API (e.g. for cases where JavaScript is disabled).

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+        response.getOutputStream().println(aboutMe);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response)
+            throws IOException {
+
+        UserService userService = UserServiceFactory.getUserService();
+        if (!userService.isUserLoggedIn()) {
+            response.sendRedirect("/index.html");
```
michaelsafyan, line 67: A redirect may make sense if this POST request is being submitted directly via a `<button type="submit">` element in an HTML form. However, it would probably be better to drive the submit operation in JavaScript (e.g. using `XMLHttpRequest`). In that case, it would be better to return an HTTP 401 "unauthenticated" error here, since it is easier to recognize as an error in client code (such as `XMLHttpRequest`) compared with a redirect (which will ulimately succeed with a 200 OK status code).

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+            return;
+        }
+
+        String userEmail = userService.getCurrentUser().getEmail();
+
+        System.out.println("Saving about me for " + userEmail);
+        // TODO: save the data
+
+        response.sendRedirect("/user-page.html?user=" + userEmail);
```
michaelsafyan, line 84: I would recommend generating a "204 No Content" response (with an empty body) or a "200 OK" where you return a JSON object that represents the saved state.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+
+        UserService userService = UserServiceFactory.getUserService();
+        if (!userService.isUserLoggedIn()) {
+            response.sendRedirect("/index.html");
+            return;
+        }
+
+        String userEmail = userService.getCurrentUser().getEmail();
+
+        System.out.println("Saving about me for " + userEmail);
```
michaelsafyan, line 80: Can you use a logging library such as [flogger](https://github.com/google/flogger) to do your logging?

src/main/webapp/user-page.html
```diff
@@ -42,5 +42,13 @@ <h1 id="page-title">User Page</h1>
 
     <div id="message-container">Loading...</div>
 
+    <div id="about-me-container">Loading...</div>
+    <div id="about-me-form" class="hidden">
```
michaelsafyan: How do you intend to retrieve the previously saved content?

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+
+        response.getOutputStream().println(aboutMe);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response)
+            throws IOException {
+
+        UserService userService = UserServiceFactory.getUserService();
+        if (!userService.isUserLoggedIn()) {
```
michaelsafyan, line 66: You can simplify this slightly by using a security constraint in the `web.xml` file:
https://cloud.google.com/appengine/docs/standard/java/config/webxml#security-auth

... to make App Engine automatically require the user to be logged-in.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
markser: I will be merging my changes to the master, take a look at the stats page if you are confused on what to do

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
markser: I agree, it might be better to have clarification on what user we are talking about in line 38.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
markser: I believe that we should have a check that a user actually exists in the database. 
I believe the current implementation has it where the field "about me" always public and visible to anyone. This is something that we can look into when we expand on who our users are and the amount of privacy we want to take in place. 
Could be cool if we had an option where they are able to click saying that they are open to having their "about me" public or not.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
markser: I believe that this will force the users to use gmail to login, is this something that we want to do?

markser (changes requested):
Just need more clarification on some lines of code. Seems like for the most part that we are asking that we want better implementation with checking the database.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
michaelsafyan: Re @markser 's feedback:

In App Engine, it is definitely easiest to just require a Google login as there is built-in support for that sort of thing. If you were creating an application where you wanted to support multiple different login options (e.g. Google, Facebook, Email+Password, etc.), you would probably want to use a solution like "Firebase Auth" (Google-owned) or "Auth0" (3rd party) to implement it (implementing authentication from scratch is not an easy task and probably not something that you ever want to do yourself, especially not without appropriate security people to vet it). In such a case, though, the code wouldn't look that dissimilar...

If you want to make this somewhat more realistic in terms of an application that supports multiple  kinds of authentication, then you could create a layer of indirection where you have your own set of authentication interfaces that wrap the App Engine auth libraries and that, in the future, could be replaced with an implementation that uses Firebase Auth or Auth0. 

For now, though, I'd recommend just learning how to use App Engine's google auth as it will provide you with an easier path to understanding these auth systems in general, and I don't think it is necessary to introduce all that complexity right away, especially if we don't ever plan to do the full blown thing in the course of this Code U (unless you want to make supporting multiple authentication mechanisms the main project that you work on down the road).

src/main/webapp/WEB-INF/web.xml
```diff
@@ -36,23 +36,33 @@
     <servlet-class>com.google.codeu.servlets.MessageServlet</servlet-class>
   </servlet>
 
+  <servlet>
+    <servlet-name>aboutme</servlet-name>
+    <servlet-class>com.google.codeu.servlets.AboutMeServlet</servlet-class>
+  </servlet>
+
   <servlet-mapping>
-    <url-pattern>/login</url-pattern>
```
michaelsafyan, line 40: What's with this reordering here?

I think it makes more sense to specify 'url-pattern' first, because from/source logically precedes to/destination...

That is, I think it is easier to read this as:

    "When you see the URL /foo, then route it to servlet bar."

Specifying the servlet name, first, rather than the URL pattern first creates a more awkward sentence for reading/understanding what the stanza does.

If you feel really strongly that it should be ordered the other way, I'm fine with that, but pull requests should be logically focused; I'd ask that if you want to reorder that you make such a change in a separate clean-up/maintenance pull request focused on this cosmetic change rather than mixing unrelated cosmetic changes into your feature change. Is that OK with you?

src/main/webapp/js/user-page-loader.js
```diff
@@ -95,4 +96,21 @@ function buildUI() {
...
   showMessageFormIfViewingSelf();
   fetchMessages();
-}
\ No newline at end of file
+  fetchAboutMe();
+}
+
+/** Fetches about me data */
+function fetchAboutMe(){
+    const url = '/about?user=' + parameterUsername;
```
michaelsafyan, line 104: Where does "parameterUsername" come from?

Here's what I'd suggest....

Organize the application using the following URLs:

      /profiles/me[?mode=VIEW|EDIT]
      /profiles/USER_NAME

In most of your code, I would make the parameters explicit (like `fetchProfileData(username)`, where the username is an explicit function parameter). Then, separately, you can have a function that explicitly retrrieves the parameter from the URL as in `extractUsernameFromProfileUrl(url)`, allowing for a fairly untuitive combination like:

```
    var url = window.location.href;
    var username = extractUsernameFromProfileUrl(url);
    fetchProfileData(username)
         .then(handleProfileDataReceived)
         .catch(handleProfileDataFetchingFailed);
```

src/main/webapp/WEB-INF/web.xml
paulinagzd: It's okay with me. I rearranged them to understand the logic and implement the changes.

src/main/webapp/js/user-page-loader.js
```diff
@@ -95,4 +96,21 @@ function buildUI() {
...
+    fetch(url).then((response) => {
+        return response.text();
+    }).then((aboutMe) => {
+        const aboutMeContainer = document.getElementById('about-me-container');
+        if(aboutMe == ''){
+            aboutMe = 'This user has not entered any information yet.';
+        }
+
+        aboutMeContainer.innerHTML = aboutMe;
```
michaelsafyan, line 114: If the user can submit arbitrary `aboutMe` content, including content that might include `<script>` tags and you assign that content to `innerHTML`, it would allow one user to inject `<script>` tags that get displayed and executed for another user, thereby allowing one user to run code that is in the security/execution context of another user.

In other words, assigning to `innerHTML` here is a potential vector for a cross-site scripting (XSS) attack:
https://en.wikipedia.org/wiki/Cross-site_scripting

This is something that even many experienced engineers do by mistake.

To mitigate this, I would recommend one of the following options:

  - Don't support rich content (assign to `textContent` which will turn HTML tags into
    the textual equivalents rather than actually treating those tags like active HTML).

  - Convert untrusted user input into trusted HTML before assigning to `innerHTML`. For example,
    you could use a markdown library that guarantees that only certain kinds of HTML will be passed
    through, with certain kinds of HTML (such as scripts) automatically escaped/removed

src/main/webapp/js/user-page-loader.js
```diff
@@ -95,4 +96,21 @@ function buildUI() {
...
-}
\ No newline at end of file
+  fetchAboutMe();
+}
+
+/** Fetches about me data */
+function fetchAboutMe(){
+    const url = '/about?user=' + parameterUsername;
+    fetch(url).then((response) => {
+        return response.text();
```
michaelsafyan, line 107: I'd recommend using `.json()` instead of `.text()`. This will make it easier to introduce more data in the response in the future. For example, let's say that right now the about me only consists of a short tagline. If you use JSON, you can have a response that looks something like:

```
{
   "tagline": "My tagline"
}
```

And then, in the future, if you want to add links to one's relevant social profiles as in:

```
{
   "tagline": "My tagline",
   "socialProfiles": [{
      "displayName": "Facebook",
      "url": "http://facebook.com/myprofile"
    }, {
      "displayName": "Twitter",
      "url": "http://twitter.com/myprofile"
    }]
}
```

... you can do so without breaking existing client code.

If, on the other hand, you put the tagline as just the raw text response, then when you add more content to the response, code that was assuming that the entire message is the tagline would be broken as a result of the update.

src/main/webapp/js/user-page-loader.js
markser: parameterUsername comes from line 19 as a FYI,
But I do like what Michael is proposing, it makes it where you just have to look around the code to understand what is going on instead of reading the entire file.

src/main/java/com/google/codeu/servlets/AboutMeServlet.java
markser: I changed the init() function to something like this
```
@Override
public void init(ServletConfig config) throws ServletException {
    super.init(config);
    datastore = new Datastore();
  }
```
This might be something that could work for you as well!

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -81,4 +81,33 @@ public void storeMessage(Message message) {
 
     return messages;
   }
+
+    /** Stores the User in Datastore. */
+    public void storeUser(User user) {
```
michaelsafyan, line 86: Consider the following scenario:

      // Request 1 (submitted by "foo@domain.tld")
      User user = ...
      storeUser(user);

      // Request 2 (also submitted by foo@domain.tld to update the "about me" data)
      User user = ....
      storeUser(user);

In this scenario, it would appear that you create duplicate entries/rows for the same user.

I would recommend, instead defining:

     public Optional<Key> getUserKeyByEmail(String email);
     public Optional<User> getUserByEmail(String email);
     public Key createOrUpdateUser(User user);

With:

     public createOrUpdateUser(User user) {
        Optional<Key> existingKeyOrAbsent = getUserKeyByEmail(user.getEmail());
        if (existingKeyOrAbsent.isPresent()) {
           Key existingKey = existingKeyOrAbsent.get();
           updateUser(existingKey, user);
           return existingKey;
         }
         return createUser(user);
     }

     private updateUser(Key key, User user) {
        // ...
     }

     private Key createUser(User user) {
        // ...
     }

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -81,4 +81,33 @@ public void storeMessage(Message message) {
...
+        Entity userEntity = new Entity("User", user.getEmail());
+        userEntity.setProperty("email", user.getEmail());
+        userEntity.setProperty("aboutMe", user.getAboutMe());
+        datastore.put(userEntity);
+    }
+
+    /**
+     * Returns the User owned by the email address, or
+     * null if no matching User was found.
```
michaelsafyan, line 95: It is best to make the possibility of null-ness/invalidity a part of the signature. To do so, switch from `User` as the return type to `Optional<User>`.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -81,4 +81,33 @@ public void storeMessage(Message message) {
 
     return messages;
   }
+
+    /** Stores the User in Datastore. */
+    public void storeUser(User user) {
+        Entity userEntity = new Entity("User", user.getEmail());
```
michaelsafyan, line 87: It's a good idea to use constants for variables like "User", "email", "aboutMe".

For example, I would recommend having something like:

     public class User {
       public static class DatastoreConstants {
          public static String ENTITY_KIND = "User";
          public static String ENTITY_EMAIL_PROPERTY = "email";
          public static String ENTITY_ABOUTME_PROPERTY = "aboutMe";
        }
        // ...
      }

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -81,4 +81,33 @@ public void storeMessage(Message message) {
...
+        Query query = new Query("User")
+                .setFilter(new Query.FilterPredicate("email", FilterOperator.EQUAL, email));
+        PreparedQuery results = datastore.prepare(query);
+        Entity userEntity = results.asSingleEntity();
+        if(userEntity == null) {
+            return null;
+        }
+
+        String aboutMe = (String) userEntity.getProperty("aboutMe");
+        User user = new User(email, aboutMe);
```
michaelsafyan, line 108: This works for now since you only have a small number of properties, but it is generally a good idea to use [the builder pattern](https://www.michaelsafyan.com/tech/design/patterns/builder) when it comes to classes that represent datastore entities. That is, I would recommend:

       User user = User
           .newBuilder()
           .setEmail(email)
           .setAboutMe(aboutMe)
           .build();

It's also a good idea to do it that way in that it makes it easier to populate only a subset of the fields (e.g. when making [projection queries](https://cloud.google.com/appengine/docs/standard/java/datastore/projectionqueries) that retrieve a partial entity).

You may want to take a look at some of the example code here:
https://cloud.google.com/java/getting-started/using-cloud-datastore

... which illustrates best practices.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -81,4 +81,33 @@ public void storeMessage(Message message) {
...
+
+        Query query = new Query("User")
+                .setFilter(new Query.FilterPredicate("email", FilterOperator.EQUAL, email));
+        PreparedQuery results = datastore.prepare(query);
+        Entity userEntity = results.asSingleEntity();
+        if(userEntity == null) {
+            return null;
+        }
+
+        String aboutMe = (String) userEntity.getProperty("aboutMe");
```
michaelsafyan, line 107: What happens here if the property hasn't been set? Does it throw? Does it produce a null value? Should there be any sort of logic to default the value to "" when the property is unset/null?

src/main/webapp/user-page.html
```diff
@@ -31,6 +31,16 @@
     </nav>
     <h1 id="page-title">User Page</h1>
 
+    <!-- This creates an about me text placeholder in the logged user's page. -->
+    <div id="about-me-container">Loading...</div>
+    <div id="about-me-form" class="hidden">
+      <form action="/about" method="POST">
+        <textarea name="about-me" placeholder="about me" rows=4 required></textarea>
+        <br/>
```
michaelsafyan, line 39: Instead of `<br/>`, I would recommend putting `<div>` elements around the pieces that you want to have on separate lines. Using `<div>` instead of `<br>` makes it easier to control the margin/padding and other aspects of the display through CSS.

In other words:

     <form ...>
      <div class="form-contents">
          <div class="input-container">
             <textarea ...>
          </div>
          <div class="input-container">
             <input ... >
          </div>
      </div>
     </form>

This will already put them on separate lines. If you want to adjust spacing between them, you can use something like the following CSS code:

       /* ... */
       .input-container {
          margin-top: 8px;
       }

       .input-container:first-child {
          margin-top: 0;
       }
       /* ... */

michaelsafyan (changes requested):
Left a few more suggestions and comments.

src/main/java/com/google/codeu/data/Datastore.java
markser: Ah, this code seems very similar to using a HashMap data structure, which is going to be very good if we want fast lookup. I believe that we can still use the current implementation and temporary circumvent the proposed scenario if we had some checks in storeUser() to see if an user is already in the database or not. 

But, I think that it would be better to use the proposed implementation because of it allows us to separate the updating and creating the user. So later down the line, we can implement different features and attributes for the users if we wanted to.

src/main/java/com/google/codeu/data/Datastore.java
markser: @michaelsafyan with using optional, should we be using Try-Catch blocks with it? Or is that redundant? 
I was thinking of having the try and catch blocks to log where we had a nullcase.

src/main/java/com/google/codeu/data/Datastore.java
michaelsafyan: Re: your feedback, @markser 

The observation that the Cloud Datastore is similar to HashMap is a good one. It does a little more than a HashMap in that it also supports indexing by multiple different entity properties (not just the key), but it is definitely similar in that each entity (row) is assigned a unique key that identifies it and that it is quick to find an entity by key (and less quick to query rows that meet some criteria).

In terms of making look-ups quick, there is a trade-off to be aware of here between consistency and speed. When storing data in a cache, there is a possibility of serving an old version from the cache after the data has been updated in the authoritative database. That issue can be mitigated with cache invalidation, but such invalidation requires centralization of the caching (rather than having each copy of the program cache in its own process's virtual memory), and that centralization, in turn, trades off some speed and availability for that consistency.

In typical real-world applications that need to speed up such lookups, you'll find that they often use a caching server (like Redis or Memcache) in conjunction with the database for caching with invalidation, expiration.

As with most optimizations, though, unless the optimization is non-trivial, such optimizations are best to defer unless/until the need for the optimization can be clearly rationalized relative to other priorities. At this point, adding Redis or Memcache to the system strikes me as premature.

src/main/java/com/google/codeu/data/Datastore.java
michaelsafyan: In terms of nullness testing (or any similar checking for that matter), the key thing to guarantee is that errors are thrown from the true source of the error, not at a later time as a result of propagated errors / invalid state.

If you use an object in the course of a function where that object's validity will be implicitly tested as a result of the use, there is no need to explicitly check the validity. For example, if you have code like:

       void doSomething(Optional<Foo> fooOrAbsent) {
             // ...
             Foo foo = fooOrAbsent.get();
             // ...
       }

In the above, there is no need to check that `fooOrAbsent` is non-null; calling `.get()` will throw `NullPointerException` (the same behavior that proper null testing that you do manually should similarly do).

If, however, you store a value that you access later in a different function as in:

       MyConstructor(Optional<Foo> fooOrAbsent) {
          // ...
          this.fooOrAbsent = fooOrAbsent;
          // ...
       }

        void doSomething() {
           // ... use fooOrAbsent here ...
        }

In a case like the above, you should validate the data before storing it. This way, you ensure the class invariant (i.e. you ensure that the state of the class is valid), and you ensure that errors are propagated in the correct location (to the code that invoked the constructor improperly, not the code that correctly invoked "doSomething").

michaelsafyan (commented):
Replied to a few of @markser's comments.

PR #11, opened 2019-03-04
Implemented direct messaging feature

shamailah:
Implemented minimum working version of the direct messaging feature. When a logged-in user visits another user's page, the messages between the logged-in user and the other user are displayed in ascending time order (most recent messages are at the bottom, as you'd expect). A logged-in user cannot view the messages between two other users. The logged-in user can currently only send messages to another user by visiting their page.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,6 +59,19 @@ limitations under the License.
       <artifactId>google-cloud-datastore</artifactId>
       <version>1.52.0</version>
     </dependency>
+
+    <dependency>
+      <groupId>com.google.flogger</groupId>
+      <artifactId>flogger</artifactId>
+      <version>0.3.1</version>
+  </dependency>
+
+  <dependency>
+      <groupId>com.google.flogger</groupId>
+      <artifactId>flogger-system-backend</artifactId>
+      <version>0.3.1</version>
+      <scope>runtime</scope>
+  </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -64,14 +64,20 @@ limitations under the License.
       <groupId>com.google.flogger</groupId>
       <artifactId>flogger</artifactId>
       <version>0.3.1</version>
-  </dependency>
+    </dependency>
 
-  <dependency>
-      <groupId>com.google.flogger</groupId>
-      <artifactId>flogger-system-backend</artifactId>
-      <version>0.3.1</version>
-      <scope>runtime</scope>
-  </dependency>
+    <dependency>
+        <groupId>com.google.flogger</groupId>
+        <artifactId>flogger-system-backend</artifactId>
+        <version>0.3.1</version>
+        <scope>runtime</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-language</artifactId>
+      <version>1.55.0</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/src/main/java/com/google/codeu/data/Datastore.java
+++ b/src/main/java/com/google/codeu/data/Datastore.java
@@ -88,32 +88,30 @@ public class Datastore {
   }
 
   /**
-   * Gets messages sent between user and a recipient.
+   * Gets messages sent between the logged-in user and another user.
    *
-   * @return a list of messages sent between the user and a recipient, or empty list if user or recipient has never sent a
-   *     message to each other. List is sorted by time ascending.
+   * @return a list of messages sent between the logged-in user and another user, or empty list
+   * if logged-in user or other user have never sent a message to each other. List is sorted by time ascending.
    */
-
-  // TODO: Finish this feature
-  public List<Message> getMessagesBetweenTwoUsers(String user, String recipient) {
+  public List<Message> getMessagesBetweenTwoUsers(String loggedInUser, String otherUser) {
     List<Message> messages = new ArrayList<>();
 
-    // Messages between two people, where user is the sender
-    Filter userSentMessages = new Query.FilterPredicate("user", FilterOperator.EQUAL, user);
-    Filter recipientReceivedMessages = new Query.FilterPredicate("recipient", FilterOperator.EQUAL, recipient);
+    // Messages between two people, where logged-in user is the sender
+    Filter messagesSentByLoggedInUser = new Query.FilterPredicate("user", FilterOperator.EQUAL, loggedInUser);
+    Filter messagesReceivedByOtherUser = new Query.FilterPredicate("recipient", FilterOperator.EQUAL, otherUser);
 
-    // Messages between two people, where user is the receiver
-    Filter userReceivedMessages = new Query.FilterPredicate("recipient", FilterOperator.EQUAL, user);
-    Filter recipientSentMessages = new Query.FilterPredicate("user", FilterOperator.EQUAL, recipient);
+    // All the messages sent by logged-in user to the other user
+    Filter loggedInUserMessages = CompositeFilterOperator.and(messagesSentByLoggedInUser, messagesReceivedByOtherUser);
 
-    // All the messages sent by user to the other person
-    Filter userMessages = CompositeFilterOperator.and(userSentMessages, recipientReceivedMessages);
-
-    // All the messaged recieved by user sent by the other person
-    Filter recipientMessages = CompositeFilterOperator.and(recipientSentMessages, userReceivedMessages);
+    // Messages between two people, where logged-in user is the recipient
+    Filter messagesSentByOtherUser = new Query.FilterPredicate("user", FilterOperator.EQUAL, otherUser);
+    Filter messagesReceivedByLoggedInUser = new Query.FilterPredicate("recipient", FilterOperator.EQUAL, loggedInUser);
+    
+    // All the messages recieved by logged-in user sent by the other user
+    Filter otherUserMessages = CompositeFilterOperator.and(messagesSentByOtherUser, messagesReceivedByLoggedInUser);
 
     // All the messages between the two users
-    Filter directMessages = CompositeFilterOperator.or(userMessages, recipientMessages);
+    Filter directMessages = CompositeFilterOperator.or(loggedInUserMessages, otherUserMessages);
 
     Query query =
         new Query("Message")
--- a/src/main/java/com/google/codeu/data/Datastore.java
+++ b/src/main/java/com/google/codeu/data/Datastore.java
@@ -53,9 +53,9 @@ public class Datastore {
   }
 
   /**
-   * Gets messages posted by a specific user.
+   * Gets messages received by a specific user.
    *
-   * @return a list of messages posted by the user, or empty list if user has never posted a
+   * @return a list of messages received by the user, or empty list if user has never received a
    *     message. List is sorted by time descending.
    */
   public List<Message> getMessages(String recipient) {
@@ -84,11 +84,6 @@ public class Datastore {
       }
     }
 
-    // test if sender/reciever messages are fetched
-    for (int i = 0; i < messages.size(); i++) {
-      System.out.println(messages.get(i));
-    }
-
     return messages;
   }
 
@@ -96,7 +91,7 @@ public class Datastore {
    * Gets messages sent between user and a recipient.
    *
    * @return a list of messages sent between the user and a recipient, or empty list if user or recipient has never sent a
-   *     message to each other. List is sorted by time descending.
+   *     message to each other. List is sorted by time ascending.
    */
 
   // TODO: Finish this feature
@@ -123,7 +118,7 @@ public class Datastore {
     Query query =
         new Query("Message")
             .setFilter(directMessages)
-            .addSort("timestamp", SortDirection.DESCENDING);
+            .addSort("timestamp", SortDirection.ASCENDING);
     PreparedQuery results = datastore.prepare(query);
 
     for (Entity entity : results.asIterable()) {
@@ -142,11 +137,6 @@ public class Datastore {
         System.err.println(entity.toString());
         e.printStackTrace();
       }
-    }
-
-    // test if sender/reciever messages are fetched
-    for (int i = 0; i < messages.size(); i++) {
-      System.out.println(messages.get(i));
     }
 
     return messages;
--- a/src/main/java/com/google/codeu/data/Datastore.java
+++ b/src/main/java/com/google/codeu/data/Datastore.java
@@ -16,6 +16,8 @@
 
 package com.google.codeu.data;
 
+import com.google.common.flogger.FluentLogger;
+import com.google.appengine.api.datastore.PropertyContainer;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
@@ -28,6 +30,7 @@ import com.google.appengine.api.datastore.Query.SortDirection;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
+import java.util.Optional;
 
 /** import for Fetch Options */
 import com.google.appengine.api.datastore.FetchOptions;
@@ -35,10 +38,25 @@ import com.google.appengine.api.datastore.FetchOptions;
 /** Provides access to the data stored in Datastore. */
 public class Datastore {
 
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private DatastoreService datastore;
 
   public Datastore() {
     datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+
+  private Optional<String> getStringProperty(PropertyContainer container, String propertyName) {
+    if (!container.hasProperty(propertyName)) {
+      return Optional.empty();
+    }
+
+    String value = null;
+    try {
+       value = (String) container.getProperty(propertyName);
+    } catch (ClassCastException wrongType) {
+        logger.atSevere().withCause(wrongType).log("Property \"" + propertyName + "\" exists but is not a String.");
+    }
+    return Optional.ofNullable(value);
   }
 
   /** Stores the Message in Datastore. */
@@ -50,41 +68,6 @@ public class Datastore {
     messageEntity.setProperty("recipient", message.getRecipient());
 
     datastore.put(messageEntity);
-  }
-
-  /**
-   * Gets messages received by a specific user.
-   *
-   * @return a list of messages received by the user, or empty list if user has never received a
-   *     message. List is sorted by time descending.
-   */
-  public List<Message> getMessages(String recipient) {
-    List<Message> messages = new ArrayList<>();
-
-    Query query =
-        new Query("Message")
-            .setFilter(new Query.FilterPredicate("recipient", FilterOperator.EQUAL, recipient))
-            .addSort("timestamp", SortDirection.DESCENDING);
-    PreparedQuery results = datastore.prepare(query);
-
-    for (Entity entity : results.asIterable()) {
-      try {
-        String idString = entity.getKey().getName();
-        UUID id = UUID.fromString(idString);
-        String text = (String) entity.getProperty("text");
-        long timestamp = (long) entity.getProperty("timestamp");
-        String user = (String) entity.getProperty("user");
-        Message message = new Message(id, user, text, timestamp, recipient);
-
-        messages.add(message);
-      } catch (Exception e) {
-        System.err.println("Error reading message.");
-        System.err.println(entity.toString());
-        e.printStackTrace();
-      }
-    }
-
-    return messages;
   }
 
   /**
@@ -113,6 +96,7 @@ public class Datastore {
     // All the messages between the two users
     Filter directMessages = CompositeFilterOperator.or(loggedInUserMessages, otherUserMessages);
 
+    // Sort in the ascending order so that most recent messages appear last
     Query query =
         new Query("Message")
             .setFilter(directMessages)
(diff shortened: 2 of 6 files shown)


src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -51,17 +53,17 @@ public void storeMessage(Message message) {
   }
 
   /**
-   * Gets messages posted by a specific user.
+   * Gets messages received by a specific user.
```
michaelsafyan: I think it might make sense for this to be the list of all messages that are visible to the given user (both those received by that user as well as messages that the user has written). Otherwise, how does the user view the messages that they have sent or posted publicly?

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -51,17 +53,17 @@ public void storeMessage(Message message) {
...
    *
-   * @return a list of messages posted by the user, or empty list if user has never posted a
+   * @return a list of messages received by the user, or empty list if user has never received a
    *     message. List is sorted by time descending.
    */
-  public List<Message> getMessages(String user) {
+  public List<Message> getMessages(String recipient) {
```
michaelsafyan: For clarity, you may want to consider renaming the function; for example "getMessagesSentTo" or "getMessagesReceivedBy", which would clarify that this function returns messages received by the given user rather than all messages which are visible to the given user. Alternatively, you may want to consider changing the definition/implementation so that this returns all messages that are visible to the given user, including both messages that the given user has sent as well as public messages (with no recipient) and those sent to this user.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -85,6 +87,59 @@ public void storeMessage(Message message) {
...
+    // All the messages recieved by logged-in user sent by the other user
+    Filter otherUserMessages = CompositeFilterOperator.and(messagesSentByOtherUser, messagesReceivedByLoggedInUser);
+
+    // All the messages between the two users
+    Filter directMessages = CompositeFilterOperator.or(loggedInUserMessages, otherUserMessages);
+
+    Query query =
+        new Query("Message")
+            .setFilter(directMessages)
+            .addSort("timestamp", SortDirection.ASCENDING);
```
michaelsafyan, line 103: Can you leave a comment explaining the choice of ASCENDING instead of DESCENDING? Is it because more recent messages may be in reply to previous messages?

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -85,6 +87,59 @@ public void storeMessage(Message message) {
...
+    PreparedQuery results = datastore.prepare(query);
+
+    for (Entity entity : results.asIterable()) {
+      try {
```
michaelsafyan, line 109: Which of these statements to do you expect to fail?

In general, a try statement should be as narrowly scoped as possible. Consider using helper functions that do a single operation within a "try" to keep the scope to a minimum. For example, you might have something like:

       private Optional<String> getStringProperty(PropertyContainer container, String propertyName) {
         if (!container.hasProperty(propertyName)) {
           return Optional.<string>absent();
         }

         String value = null;
         try {
           value = (String) container.getProperty(propertyName);
         } catch (ClassCastException wrongType) {
            logging.atSevere().withCause(wrongType).log("Property \"" + propertyName + "\" exists but is not a String.");
         }
         return Optional.<String>ofNullable(value);
     }

... assuming that this is the aspect of the code below you expect might fail.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -85,6 +87,59 @@ public void storeMessage(Message message) {
...
+        String idString = entity.getKey().getName();
+        UUID id = UUID.fromString(idString);
+        String text = (String) entity.getProperty("text");
+        long timestamp = (long) entity.getProperty("timestamp");
+        String sender = (String) entity.getProperty("user");
+        String receiver = (String) entity.getProperty("receiver");
+        Message message = new Message(id, sender, text, timestamp, receiver);
+
+        messages.add(message);
+      } catch (Exception e) {
```
michaelsafyan: This kind of general catch statement for debug logging is OK at higher-levels in the stack (e.g. if you want to create a `ServletBase` class that all the servlets in this application extend and that injects this kind of generic error-handling for uncaught exceptions), but is a pattern that is best avoided otherwise. Other than in framework code high up in the stack, functions should generally catch narrowly-defined exception type(s).

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -40,23 +40,29 @@ public void init() {
...
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
     response.setContentType("application/json");
 
-    String user = request.getParameter("user");
-
-    if (user == null || user.equals("")) {
+    // Get logged-in user and other user
+    String loggedInUser = null;
```
michaelsafyan: I would suggest using `Optional<String>` for `loggedInuser` and `otherUser`.

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -40,23 +40,29 @@ public void init() {
...
+    String otherUser = request.getParameter("user");
+    UserService userService = UserServiceFactory.getUserService();
+    if (userService.isUserLoggedIn()) {
+      loggedInUser = userService.getCurrentUser().toString();
```
michaelsafyan: Can `getCurrentUser()` return null? If so, it looks like this code will throw `NullPointerException`.

michaelsafyan: Based on the documentation here:
https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/User

It is unclear what `toString()` will return and whether there are any guarantees on the content/stability of the function. Based on the lack of promises/guarantees, it would seem that `toString()` should return something useful for debugging but may not be a stable API and may not return what you expect/want in the future.

I would recommend making this code more reliable and clear by explicitly retrieving a property that you care about such as `.getEmail()` or `.getUserId()`, depending on how you are identifying the user.

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -40,23 +40,29 @@ public void init() {
...
+    }
+    
+    if (loggedInUser == null || loggedInUser.equals("") || otherUser == null || otherUser.equals("")) {
```
michaelsafyan: For error cases, the common practice in web services is to set an HTTP status code that indicates an error.

There is a list of HTTP status codes here:
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

For the case where the user is not logged-in, the expected code would be:
401 ("Unauthenticated")

For the case where the required GET parameter is absent, the expected code would be:
400 ("Bad Request"), which is logically equivalent to "InvalidArgumentException"

You can set that status code with the `setStatus` function as in:

     if (!isValidUser(loggedInUser)) {
          response.setStatus(401);
          response.getWriter().println(...);
          return;
     }
     if (!isValidUser(otherUser)) {
          response.setStatus(400);
          response.getWriter().println(...);
          return;
     }

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -40,23 +40,29 @@ public void init() {
...
       // Request is invalid, return empty array
       response.getWriter().println("[]");
```
michaelsafyan: In error cases, it is typical to return a differently-structured response that conveys relevant error information.

For example:

      {
          "error": {
             "message": "Missing required query parameter 'user'"
          }
      }

For how Google handles errors in its own APIs, see:
https://cloud.google.com/apis/design/errors

michaelsafyan (changes requested):
Assuming that you've tested this and it works, I'm OK with you submitting this once the general "catch" has been moved higher-up in the stack or removed, though I would recommend attending to some of the other feedback, as well, if not in this pull request then in a follow-up pull request.

src/main/java/com/google/codeu/data/Datastore.java
shamailah: Okay I changed the naming of the function to make it more intuitive. I kept the logic of the function the same though because this is intended to display all the messages a person has received on their page, kind of like a discussion board. This function doesn't get used though since I'm using the function "getMessagesBetweenTwoUsers" instead to implement direct messaging, so maybe I should just delete this function entirely?

src/main/java/com/google/codeu/data/Datastore.java
shamailah: I added a comment, but I chose ascending order so that the most recent messages appear last/bottom of a list, like what you see in a typical thread or chat.

src/main/java/com/google/codeu/data/Datastore.java
michaelsafyan: If the function is not used, then I would simply delete it.

src/main/java/com/google/codeu/data/Datastore.java
michaelsafyan: I don't see the comment. Have you pushed the change? (I'm talking about a comment just before the `.addSort`, not necessarily a comment on the function as a whole.

shamailah: I haven't pushed the changes yet! I'm just replying to your comments as I make changes, so I know I've done them and then once I make all the changes, I'll push them. Sorry for the confusion.

michaelsafyan: No worries. In the future, my recommendation would be to push changes, first, and then resolve comments. The comment resolving notifies the reviewer(s) that they should take another look.

src/main/java/com/google/codeu/servlets/MessageServlet.java
shamailah: I'm confused about this. I don't think that `loggedInUser` or `otherUser` should be optional, so would using `Optional<String>` imply that I can not pass in one of them into the function? Because if that were the case, then you wouldn't be able to see any messages at all.

src/main/java/com/google/codeu/data/Datastore.java
shamailah: Got it!

src/main/java/com/google/codeu/servlets/MessageServlet.java
michaelsafyan: The `Optional` type doesn't necessarily mean that it is valid for the object to be absent, just as the use of a nullable type doesn't necessarily mean that `null` is a legal/permissible value.

Rather, the use of `Optional` simply draws attention to the fact that absence is a possibility and is something that needs to be taken into consideration. It is common practice in many Java code bases to assume that anything which is not using the `Optional` type or that is not annotated with the `@Nullable` type annotation will be non-null (and thus checking for null is unnecessary).

When you use:

     String user = ...

It makes it less obvious that code below must test that `user != null`.

If you, instead, write:

     Optional<String> user = ...

... then it makes it more obvious that the code below must check that the value is present.

You can always assign to a pure, non-optional type once the check has been applied as in:

      Optional<String> userOrAbsent = ...
      if (userOrAbsent.orElse("").isEmpty()) {
         // ... generate an error ...
        return;
      }
      String user = userOrAbsent.get();

src/main/java/com/google/codeu/data/Datastore.java
shamailah: I'm not sure I understand this

src/main/java/com/google/codeu/data/Datastore.java
shamailah: What do you mean by keeping the try scope to minimum? Only check a small number of things within a try? So if I expected all the things within the try statement, as it is right now, to fail, I should have a separate try catch for every property?

src/main/java/com/google/codeu/data/Datastore.java
shamailah: What would be a better catch statement?

src/main/java/com/google/codeu/data/Datastore.java
michaelsafyan: Yep, that's exactly right; the number of operations performed inside a "try { ... }" should be the minimum number of statements possible. To keep this code from being ugly with lots of separate "try { ... }" blocks, I would use a subroutine (or multiple subroutines) such as the "getStringProperty" function that I prototyped above, where the subroutine provides for a path to graceful error handling.

For example, with the "getStringProperty" code that I suggested above, you could do something like:

       String text = getStringProperty(entity, "text").orElse("");

... which causes an empty string to be used as a fallback in the case of an error.

src/main/java/com/google/codeu/data/Datastore.java
michaelsafyan: If you want to do this kind of general catch for debugging, I would simply move it higher-up in the stack; that is, there is a function called `service`:

https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/http/HttpServlet.html#service(javax.servlet.http.HttpServletRequest,%20javax.servlet.http.HttpServletResponse)

... that is responsible for calling `doGet`, `doPost`, etc. 

I would do something like defining:

     com.google.appengine.api.utils.SystemProperty;

     class CodingWithClassServletBase extends HttpServlet {
         @Override
          protected void service(HttpServletRequest req, HttpServletResponse resp)
                throws ServletException, java.io.IOException {
             try {
               super.service(req, resp);
             } catch (Exception e) {
               if (isDebugMode()) {
                  // When running locally, apply special error handling to make it
                  // easier to see the exception that was thrown.
                   e.printStackTrace();
                   resp.setStatus(500);
               } else {
                   // In the production environment, use the standard built-in error handling.
                   throw e;
               }
          }

          private boolean isDebugMode() {
            return SystemProperty.environment.value() != SystemProperty.Environment.Value.Production;
          }
      }

... and then I would have this class extend `CodingWithClassServletBase` instead of `HttpServlet`.

If you have exception handling that is specific to `getMessagesBetweenTwoUsers`, then you should catch specific exception types (e.g. `ClassCastException`, `IOException`, etc. as appropriate) and not the `Exception` base class.

michaelsafyan (commented):
Left some responses to the questions on the prior round of review. Hope that clarifies things. Let me know if you still have any questions.

src/main/java/com/google/codeu/data/Datastore.java
shamailah: What is `Optional.<string>absent()?`  I don't see that in the Java documentation

michaelsafyan: Sorry for the mistake/confusion. The `Optional` type was first introduced in Guava (Google's Java core libraries) before it became standardized as part of Java 8; when it was incorporated in Java 8, there were some minor naming changes to some of the methods.

I meant to say `Optional.<String>empty`:
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#empty--

`Optional.absent` is the legacy name in Guava that isn't available in Java 8:
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Optional.html#absent()

shamailah: It seems like I can't use the Google FluentLogger in the code? I think the import for it isn't in the codeu package.

michaelsafyan: You will need to update the `pom.xml` file to add a dependency on Flogger and add the relevant import statements for it to work.

The repository page has the instructions:
https://github.com/google/flogger

You can see the available `<dependencies>` that can be added in the `pom.xml` file here:
https://mvnrepository.com/artifact/com.google.flogger

If you click through the links to a specific version (you need `flogger` and `flogger-system-backend`), it will give you a `<dependency>...</dependency>` XML snippet that you can copy+paste into the `pom.xml` file.

src/main/java/com/google/codeu/data/Datastore.java
shamailah: Alright, I think I was able to implement the changes you suggested. Let me know what you think

pom.xml
```diff
@@ -59,6 +59,19 @@ limitations under the License.
       <artifactId>google-cloud-datastore</artifactId>
       <version>1.52.0</version>
     </dependency>
+
+    <dependency>
+      <groupId>com.google.flogger</groupId>
+      <artifactId>flogger</artifactId>
+      <version>0.3.1</version>
+  </dependency>
```
michaelsafyan: Minor nitpick: can you maintain consistent indentation here?

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -113,28 +96,31 @@ public void storeMessage(Message message) {
     // All the messages between the two users
     Filter directMessages = CompositeFilterOperator.or(loggedInUserMessages, otherUserMessages);
 
+    // Sort in the ascending order so that most recent messages appear last
```
michaelsafyan, line 99: Thanks for adding the comment. One additional piece of feedback related to this:

In terms of making comments valuable, focus on _why_ something is being done, not _what is being done_. That "most recent messages appear last" describes what is happening, but not why you want it to happen.

I think it would make be more helpful to do:

      Query query =
         new Query("Message")
            .setFilter(directMessages)
            // Messages are ordered from oldest to newest since messages between
            // people may include implicit context/reference to previously sent messages,
            // and this order makes the chronology of the conversation and context
            // more intuitive and understandable in the messages list page. 
            .addSort("timestamp", SortDirection.ASCENDING);

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -49,20 +50,26 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     response.setContentType("application/json");
 
     // Get logged-in user and other user
-    String loggedInUser = null;
-    String otherUser = request.getParameter("user");
+    Optional<String> loggedInUser = Optional.ofNullable(null);
```
michaelsafyan, line 53: Simplify: `Optional.ofNullable(null)` -> `Optional.empty()`

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -113,28 +96,31 @@ public void storeMessage(Message message) {
...
-      } catch (Exception e) {
+        idString = entity.getKey().getName();
+      } catch (NullPointerException e) {
         System.err.println("Error reading message.");
         System.err.println(entity.toString());
         e.printStackTrace();
       }
+
+      UUID id = UUID.fromString(idString);
+      String text = (String) getStringProperty(entity, "text").orElse("");
```
michaelsafyan, line 118: The cast is no longer necessary.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -113,28 +96,31 @@ public void storeMessage(Message message) {
...
+      long timestamp = (long) entity.getProperty("timestamp");
```
michaelsafyan, line 119: Can you create a similar `getLongProperty` method?

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -113,28 +96,31 @@ public void storeMessage(Message message) {
...
     Query query =
         new Query("Message")
             .setFilter(directMessages)
             .addSort("timestamp", SortDirection.ASCENDING);
     PreparedQuery results = datastore.prepare(query);
 
     for (Entity entity : results.asIterable()) {
+
+      String idString = null;
       try {
```
michaelsafyan, line 109: When can `entity` or `entity.getKey()` be null? To my understanding, this is a situation that should not be possible (every entity that has been successfully retrieved from the database has a non-null key).

I would recommend simplifying the entire `try`...`catch` block you have here to:

     String idString = entity.getKey().getName();

michaelsafyan (commented):
Thanks! This is really great progress. I've left a few more pieces of feedback, but this is getting really close to being complete. Keep up the good work!

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -40,23 +41,35 @@ public void init() {
...
+    Optional<String> otherUser = Optional.ofNullable(request.getParameter("user"));
+    UserService userService = UserServiceFactory.getUserService();
+    if (userService.isUserLoggedIn()) {
+      loggedInUser = Optional.ofNullable(userService.getCurrentUser().getEmail());
+    }
+    
+    // Request is invalid, return empty array
+    if (loggedInUser.orElse("").isEmpty()) {
+      response.setStatus(401);
+      response.getWriter().println("Error: Unauthorized access");
```
michaelsafyan, line 63: I would strongly recommend making the error response a JSON object. That will allow you to add more properties/information to the error in the future, but that change can be made in a separate pull request.

src/main/java/com/google/codeu/data/Datastore.java
shamailah: Is this necessary? Since the queries are ordered by timestamp earlier in the method, at this point should we assume that a timestamp does in fact exist? If it didn't, would the earlier query return an error?

shamailah: This is also why I didn't use the `getStringProperty` method for retrieving the `user` and `recipient` because the earlier query would ensure that all the messages had the user and recipient I'm looking for.

michaelsafyan: That's a fair point. No, it isn't necessary.

michaelsafyan (approved):

PR #12, opened 2019-03-05
Followed up the suggestions on StatsPage

markser:
I believe that I was able to implement all of the suggestions that was provided on improving the StatsPage. I just think that I wasn't able to fully understand "loading the JavaScript and issuing this fetch" for my stats.html
I believe I was able to implement it, but I don't really understand what I did / I don't think that I did it correctly.

### DIFF
--- a/src/main/webapp/js/stats-loader.js
+++ b/src/main/webapp/js/stats-loader.js
@@ -1,23 +1,22 @@
-// Fetch stats and display them in the page.
-function fetchStats(){
-    const url = '/stats';
-    fetch(url).then((response) => {
+function fetchStats() {
+  const url = "/stats";
+  fetch(url)
+    .then(response => {
       return response.json();
-    }).then((stats) => {
-      const statsContainer = document.getElementById('stats-container');
-      statsContainer.innerHTML = '';
-      const messageCountElement = buildStatElement('Message count: ' + stats.messageCount);
+    })
+    .then(stats => {
+      const statsContainer = document.getElementById("stats-container");
+      statsContainer.innerHTML = "";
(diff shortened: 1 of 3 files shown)


src/main/webapp/js/stats-loader.js
```diff
@@ -1,23 +1,22 @@
...
+  fetch(url)
+    .then(response => {
       return response.json();
-    }).then((stats) => {
-      const statsContainer = document.getElementById('stats-container');
-      statsContainer.innerHTML = '';
-      const messageCountElement = buildStatElement('Message count: ' + stats.messageCount);
+    })
+    .then(stats => {
+      const statsContainer = document.getElementById("stats-container");
```
michaelsafyan: It's possible that the document will not have completed loading by this point, and thus the `stats-container` might not yet exist. I'd suggest the following minor refactor:


     var _statsPromise = null;
     function getOrPrefetchStats() {
       if (_statsPromise == null) {
         _statsPromise = fetch(url).then(response => { return response.json(); });
       }
       return _statsPromise;
     }

     function renderStatsOnPage() {
       getOrPrefetchStats().then(stats = > {
            const statsContainer = ...
            ...
       });
     }

And then in the HTML:

    ...
    <head>
       ...
     <script>getOrPrefetchStats();</script>
    </head>
    <body onload="renderStatsOnPage">
        ...
    </body>

michaelsafyan (approved):
Approving on the condition/assumption that you will address the one remaining comment.

src/main/webapp/js/stats-loader.js
markser: Perfect, this makes sense to me now!

PR #13, opened 2019-03-06
Starter Code for Chart Page

markser:
This is the starter code for the Chart Page. All data right now will be static for right now and needs some clean up.
This will display a bar chart with /chart.html

### DIFF
--- a/src/main/webapp/chart.html
+++ b/src/main/webapp/chart.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+
+<html>
+  <head>
+    <script
+      type="text/javascript"
+      src="https://www.gstatic.com/charts/loader.js"
+    ></script>
+    //TODO move all of this to a separate JavaScript file
+    //figure out a way to get user input/data for the charts
+    //make sure to implement same idea as in the stats page
+    <script type="text/javascript">
+      google.charts.load("current", { packages: ["corechart"] });
+      google.charts.setOnLoadCallback(drawChart);
+      function drawChart() {
+        var book_data = new google.visualization.DataTable();
+        //define columns for the DataTable instance
+        book_data.addColumn("string", "Book Title");
+        book_data.addColumn("number", "Votes");
+
+        //add data to book_data
+        book_data.addRows([
+          ["The Best We Could Do", 6],
+          ["Sing, Unburied, Sing", 10],
+          ["The Book of Unknown Americans", 7],
+          ["The 57 Bus", 4],
+          ["The Handmaid's Tale", 8]
+        ]);
+        var chart = new google.visualization.BarChart(
+          document.getElementById("book_chart")
+        );
+
+        //optional customization as dictionary
+        var chart_options = {
+          width: 800,
+          height: 400
+        };
+        chart.draw(book_data, chart_options);
+      }
+    </script>
+    <title>A New Chart</title>
+  </head>
+
+  <body>
+    <h1>Chart Project</h1>
+    <div id="book_chart"></div>
+  </body>
+</html>
--- a/src/main/webapp/chart.html
+++ b/src/main/webapp/chart.html
@@ -1,48 +1,16 @@
 <!DOCTYPE html>
 
 <html>
-  <head>
-    <script
-      type="text/javascript"
-      src="https://www.gstatic.com/charts/loader.js"
-    ></script>
-    //TODO move all of this to a separate JavaScript file
-    //figure out a way to get user input/data for the charts
-    //make sure to implement same idea as in the stats page
-    <script type="text/javascript">
-      google.charts.load("current", { packages: ["corechart"] });
-      google.charts.setOnLoadCallback(drawChart);
-      function drawChart() {
-        var book_data = new google.visualization.DataTable();
-        //define columns for the DataTable instance
-        book_data.addColumn("string", "Book Title");
-        book_data.addColumn("number", "Votes");
 
-        //add data to book_data
-        book_data.addRows([
-          ["The Best We Could Do", 6],
-          ["Sing, Unburied, Sing", 10],
-          ["The Book of Unknown Americans", 7],
-          ["The 57 Bus", 4],
-          ["The Handmaid's Tale", 8]
-        ]);
-        var chart = new google.visualization.BarChart(
-          document.getElementById("book_chart")
-        );
+<head>
+  <script src="https://www.gstatic.com/charts/loader.js"></script>
+  <script src="/js/chart-loader.js"></script>
+  <title>A New Chart</title>
+</head>
 
-        //optional customization as dictionary
-        var chart_options = {
-          width: 800,
-          height: 400
-        };
-        chart.draw(book_data, chart_options);
-      }
-    </script>
-    <title>A New Chart</title>
-  </head>
+<body>
+  <h1>Chart Project</h1>
+  <div id="bookChart"></div>
+</body>
 
-  <body>
-    <h1>Chart Project</h1>
-    <div id="book_chart"></div>
-  </body>
 </html>
--- a/src/main/webapp/chart.html
+++ b/src/main/webapp/chart.html
@@ -10,7 +10,10 @@
 
 <body>
   <h1>Chart Project</h1>
-  <div id="diffTeaChart"></div>
+  <div class="flex-container">
+    <div id="caffeine_chart"></div>
+    <div id="tea_chart"></div>
+    <div id="pie_chart"></div>
 </body>
 
 </html>
--- a/src/main/webapp/chart.html
+++ b/src/main/webapp/chart.html
@@ -10,7 +10,7 @@
 
 <body>
   <h1>Chart Project</h1>
-  <div id="bookChart"></div>
+  <div id="diffTeaChart"></div>
 </body>
 
 </html>
--- a/src/main/webapp/js/chart-loader.js
+++ b/src/main/webapp/js/chart-loader.js
@@ -1,0 +1,31 @@
+
+google.charts.load("current", {packages: ["corechart"],});
+google.charts.setOnLoadCallback(
+  function() {
+    drawChartForDiffTypesOfTea();
+  });
+
+(function drawChartForDiffTypesOfTeaForDiffTypesOfTea() {
+  var diffTypeOfTea = new google.visualization.DataTable();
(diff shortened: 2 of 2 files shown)


src/main/webapp/chart.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html>
+
+<html>
+  <head>
+    <script
+      type="text/javascript"
```
michaelsafyan: Minor note: the `type` tag is no longer required in HTML5; if `type` is unset, it defaults to `text/javascript`.

src/main/webapp/chart.html
```diff
@@ -0,0 +1,48 @@
...
+      src="https://www.gstatic.com/charts/loader.js"
+    ></script>
+    //TODO move all of this to a separate JavaScript file
+    //figure out a way to get user input/data for the charts
+    //make sure to implement same idea as in the stats page
+    <script type="text/javascript">
+      google.charts.load("current", { packages: ["corechart"] });
```
michaelsafyan: I recommend wrapping your JavaScript code in a function to prevent leakage of symbols into the global scope as in:

    (function(){
       var drawChart = function() {
          // ...
       };
 
       var init = function() {
         google.charts.load('current', {  packages: ['corechart'] });
         google.charts.setOnLoadCallback(drawChart);
       };

       init();
    })();

The outer `function(){ ... }` creates a scope that prevents the symbols from becoming global, while the `();` at the end invokes the anonymous function.

src/main/webapp/chart.html
```diff
@@ -0,0 +1,48 @@
...
+      google.charts.setOnLoadCallback(drawChart);
+      function drawChart() {
+        var book_data = new google.visualization.DataTable();
```
michaelsafyan: Minor feedback: in JavaScript, it is more common to see `camelCase` than `lower_case_with_underscores` for variable names. See, for example:

https://google.github.io/styleguide/jsguide.html#naming-local-variable-names

I'm personally not a huge stickler for these conventions (as long as we are self-consistent), but I was instructed that Code U projects should uphold the Google style guides (I believe the purpose is to teach how to adhere to style guides as in a real software company's project, even when those style rules can sometimes be grating). Given that, I would appreciate it if you followed those conventions here.

michaelsafyan (changes requested):
I think it's probably a good idea to make this a little more realistic before committing if you don't mind. I think it's fine to generate fake data (like have a function that returns a `Promise` that, instead of using `fetch`, simply generates/calculates the data), but I think your chart should be topical to what you actually want to accomplish, like, for example graphing caffeine over time.

I think it's also worth thinking about your URL structure here...

For example, I think there are probably two different kinds of pages:

     /stats   <= statistics global across all users
     /users/[USER_NAME]/stats  <= statistics about a particular user

It's unclear to me that there should be separate `stats` and `chart` pages. Do you want to clarify what URL structure you had in mind for this feature?

src/main/webapp/chart.html
markser: I blame the https://sites.google.com/codeustudents.com/spring-2019/weeks-3-7-guided-projects/charts-part-1 for not using camel case. :D 
Sounds good, I'll make sure to read through the Style Guide.

markser:
I am currently thinking of having the charts and stats page separate as of right now so that I can quickly test them separately. But I think that I would have a user specific stats page where they can answer a daily form and show them how much they are drinking (in gallons), favorite tea, amount of caffeine, leader board for the most amount of tea in the week, but I have to think of a way to incorporate user privacy while creating the leader board.

src/main/webapp/js/chart-loader.js
```diff
@@ -0,0 +1,31 @@
+
```
markser, line 40: I have lines 2 - 7 outside because we only need to call line 2 and 3 once. Let me know if I'm wrong, but I think we are fine with having them in the global scope.

markser: nvm, I'm dumb.

src/main/webapp/chart.html
```diff
@@ -1,48 +1,16 @@
...
-          height: 400
-        };
-        chart.draw(book_data, chart_options);
-      }
-    </script>
-    <title>A New Chart</title>
-  </head>
+<body>
+  <h1>Chart Project</h1>
+  <div id="bookChart"></div>
```
michaelsafyan: Minor nit: although it is standard in JavaScript to use `camelCase`, it is more common in CSS identifiers to use `lower-case-with-hyphens` for `id` and `class` attributes.

src/main/webapp/js/chart-loader.js
michaelsafyan: You are smart. Everyone gets confused. There is no need to disparage oneself.

src/main/webapp/js/chart-loader.js
```diff
@@ -0,0 +1,31 @@
...
+google.charts.load("current", {packages: ["corechart"],});
+google.charts.setOnLoadCallback(
+  function() {
```
michaelsafyan: It isn't necessary to add the extra layer of `function()` here. I would just reference the function by name.

That is:

    google.charts.setOnLoadCallback(callbackName);

Not:

    google.charts.setOnLoadCallback(function() {
      callbackName();
    });

There is no difference between the two statements logically, and the first one is simpler.

src/main/webapp/js/chart-loader.js
```diff
@@ -0,0 +1,31 @@
...
+    drawChartForDiffTypesOfTea();
```
michaelsafyan: This function is no longer visible/in-scope. That is why I recommended doing the `google.charts.setOnLoadCallback` within the scope of the anonymous function in which the callback function is defined.

src/main/webapp/js/chart-loader.js
```diff
@@ -0,0 +1,31 @@
...
+  var chartOptions = {
+    'title': 'Different Types of Tea We Drink in a Day',
+    width: 800,
+    height: 400
+  };
+  var chart = new google.visualization.BarChart(
+    document.getElementById("bookChart")
+  );
+  chart.draw(diffTypeOfTea, chartOptions);
+})();
```
michaelsafyan, line 96: By calling this here, you are causing the function to be invoked immediately, even before `google.charts.load` finishes loading the JavaScript dependencies. That may result in the symbols you use (e.g. `google.visualization.DataTable`, `google.visualization.BarChart`) not yet being valid / existing when this runs and causing an error.

src/main/webapp/js/chart-loader.js
```diff
@@ -0,0 +1,31 @@
...
+    ["White Tea", 8]
+  ]);
+  //cutomization to the chart
+  var chartOptions = {
+    'title': 'Different Types of Tea We Drink in a Day',
+    width: 800,
+    height: 400
+  };
+  var chart = new google.visualization.BarChart(
+    document.getElementById("bookChart")
```
michaelsafyan: "book" -> "tea"

src/main/webapp/js/chart-loader.js
michaelsafyan: I suggest reading these links to better understand what I'm requesting:

- [JavaScript Encapsulation With Anonymous Functions](https://www.codeproject.com/Articles/757949/JavaScript-Encapsulation-using-Anonymous-Functions)
- [What is the purpose in wrapping...?](https://stackoverflow.com/questions/2421911/what-is-the-purpose-of-wrapping-whole-javascript-files-in-anonymous-functions-li)

Note that in normal JavaScript development, *.js files aren't typically used directly; rather, a development *.js or *.ts file is compiled to a production *.js file that is actually used in serving in the web browser (see [Babel](https://babeljs.io/]), [Closure Compiler](https://developers.google.com/closure/compiler/), and TypeScript](https://www.typescriptlang.org/) for more info). This compilation process often automatically adds these anonymous function wrappers (`(function(){ ... })();`) to the files so that it does not need to be done manually. This compilation process will also typically [minify](https://en.wikipedia.org/wiki/Minification_(programming)) the JavaScript.

However, in our code base, we are not compiling the *.js files but rather using them directly. Since they are being used directly, it is useful to add these `(function(){ ... })();` wrappers manually. Although each *.html file currently uses its own, unique *.js file (and, therefore, it isn't at the moment super important), having this kind of encapsulation of the symbols of each file begins to become important when mixing *.js files together on multiple pages.

src/main/webapp/js/chart-loader.js
markser: Currently the implementation doesn't make sense when I am only using one chart, but when using multiple charts, I think that this implementation is better.
```
google.charts.setOnLoadCallback(function() {
  callbackName1();
  callbackName2();
  callbackName3();
});
```
than this:
```
google.charts.setOnLoadCallback(callbackName1);
google.charts.setOnLoadCallback(callbackName2);
google.charts.setOnLoadCallback(callbackName3);
```

src/main/webapp/js/chart-loader.js
markser: Yup, this is fixed in the most recent commit : https://github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9/pull/13/commits/04e4fb13553802490af0d69715e7e2c6a18d0881
This is why I updated my commit to I'm dumb :P

michaelsafyan (approved):

src/main/webapp/js/chart-loader.js
michaelsafyan: That's fine, though why not:

     google.charts.setOnLoadCallback(chartsLibraryLoaded);

With:

     var chartsLibraryLoaded = function() {
       callback1();
       callback2();
       callback3();
     };

I don't see why it has to be an anonymous inline function. That being said, this is minor stylistic point, and either approach is valid.

src/main/webapp/chart.html
michaelsafyan: Please address this one last comment before submitting.

markser: updated CSS identifiers to use lower-case-with-hyphens

markser:
Added more charts to the charts page
```
TODO:
Change the chart static data to dynamic data. 
Maybe a form in JavaScript for the users to input. 
```
Current plan:
- 1st work globally where everyone can input their data into the charts
- 2nd focus on user specific chart
- 3rd merge stats and charts together.

PR #15, opened 2019-03-11
Getting familiar with CSS: minor initial theme change

paulinagzd:


src/main/webapp/css/main.css
```diff
@@ -14,8 +14,20 @@
  * limitations under the License.
  */
 
+/*
+  Setting our title theme */
+title {
```
michaelsafyan: It is normally preferable to select elements by a class rather than by a tag; matching by tag often leads to overmatching, which then results in additional CSS to undo the overmatching rules. Can you match the element(s) by a class attribute (adding one if needed in the corresponding HTML) rather than matching all `title` tags?

src/main/webapp/css/main.css
```diff
@@ -14,8 +14,20 @@
...
+    font-family: Futura-PT;
```
michaelsafyan: When it comes to fonts in CSS, you either need to use a font that is common/generic enough that all browsers support it, include code that will download the font if it is not installed, or specify fall back fonts that should be used if the font isn't available (or a combination of the above ... code to download plus a fallback font).

As Futura PT appears to be a proprietary Adobe font, it is unlikely that this will work on any computer except for ones that have the Adobe Creative Suite installed and that have installed this particular font (unless you make the font available for download on the server and include an appropriate `<link>` tag to download it, something which you should not do when it comes to using a public, open source repository such as this one).

To use a fallback, you can do something like:

     font-family: Futura-PT, sans-serif;

... which essentially uses a generic sans-serif font if Futura-PT is not available.

If you're looking at customizing font, you may want to take a look at [Google Fonts](https://fonts.google.com); not only are those fonts free to use, but the website will provide you with the necessary `<link>` tag that you need to add in the `<head>` section of the corresponding `<html>` page in order to cause the font to be downloaded.

src/main/webapp/css/main.css
```diff
@@ -14,8 +14,20 @@
...
+    font-stretch: condensed;
+    font-kerning: normal;
+    text-align: right;
+    /* background-image: url("https://mir-s3-cdn-cf.behance.net/project_modules/1400/a4b55761858317.5a7c23244e1b0.jpg") ;
```
michaelsafyan: Can you delete commented-out code?

src/main/webapp/css/main.css
```diff
@@ -14,8 +14,20 @@
...
+    background-attachment: scroll ;
+    */
+}
 /* Remove default styling from the body tag. */
 body {
+    background-color: cornsilk;
```
michaelsafyan: In order to ensure that the color is interpreted consistently across different browsers, OSes, etc. it is preferable to refer to colors using RGB or RGBA format (often expressed in hex like `#RRGGBB` or using `rgb(red_val, green_val, blue_val)` or `rgba(red_val, green_val, blue_val, alpha_val)`).

For example, the color black can be expressed as `#000000` (which can be simplified to `#000`) or as `rgb(0, 0, 0)` or as `rgba(0, 0, 0, 1)`. By contrast, white can be expressed as `#ffffff` (which can be simplified to `#fff`) or as `rgb(255, 255, 255)` or `rgba(255, 255, 255, 1)`. The highest intensity red can be expressed as `#ff0000` (or simply `#f00`) or `rgb(255, 0, 0)`.

michaelsafyan (changes requested):
I've left a few comments. After addressing them, can you provide a screenshot of what things look like?

src/main/webapp/css/main.css
```diff
@@ -28,10 +31,19 @@ body {
 
 /* Make the main content look like a card. */
 .content {
-    border: 1px solid black;
+    border:# 2px solid black;
```
michaelsafyan: It looks like the '#' was accidentally put on the wrong line?

src/main/webapp/css/main.css
```diff
@@ -14,26 +14,68 @@
...
-  padding: 0;
-  margin: 0;
+/* Title animation*/
+.border{
+    background: linear-gradient(110deg, #EBCB8C 33%, rgba(0, 0, 0, 0) 33%), linear-gradient(110deg, #ffb972 34%, #88D8B0 34%);
+    background-size: 300% 300%;
+    background-position: 25% 50%;
+    padding: 5px;
+    height: 300px;
+    -webkit-animation: Gradient 15s ease infinite;
```
michaelsafyan: It should be possible to use this unprefixed:
https://developer.mozilla.org/en-US/docs/Web/CSS/animation

src/main/webapp/css/main.css
```diff
@@ -14,26 +14,68 @@
...
+    -moz-animation: Gradient 5s ease infinite;
 }
 
-/* Limit the content width and horizontally center it . */
+/* Default or base title attributes */
+.title {
+    vertical-align: middle;
+    text-align: left;
+    font-family: Futura-PT, sans-serif;
```
michaelsafyan: Did you mean to change this to 'Lato'?

src/main/webapp/css/main.css
```diff
@@ -14,26 +14,68 @@
...
+    font-kerning: normal;
+    font-size:50pt;
```
michaelsafyan: Please be consistent with spacing (always have exactly one space after ':')

src/main/webapp/css/main.css
```diff
@@ -14,26 +14,68 @@
...
+/* animation sliding */
+.title{
+    animation: Gradient 5s ease infinite;
+    -webkit-animation: Gradient 15s ease infinite;
+    -moz-animation: Gradient 5s ease infinite;
+}
+
+@-webkit-keyframes Gradient {
+    0% {
+        background-position: 30% 50%
```
michaelsafyan: Please use semicolons consistently.

src/main/webapp/css/main.css
```diff
@@ -14,26 +14,68 @@
...
 }
 
-/* Make the main content look like a card. */
+/* Text styling and alignment */
 .content {
-    border: 1px solid black;
-    margin-top: 15px;
-    padding: 16px;
-    border-radius: 8px;
+    font-family: Helvetica;
```
michaelsafyan: "Helvetica" -> "Helvetica, sans-serif" so that there is a fallback if the font isn't available.

michaelsafyan (approved):
Just a handful of comments. I'll let you commit back into the main repository whenever you feel like it is acceptable, though.

src/main/webapp/css/main.css
shamailah: Perhaps add a comment explaining the colors you're using.

shamailah (approved):
Overall looks good! Just a few stylistic changes.

michaelsafyan (approved):

PR #16, opened 2019-03-12
Created cloudbuild.yaml for cloudbuild

markser:
I created the cloudbuild.yaml for cloud build. Hopefully this is what we need to get it up and running. We might need a docker file. However, I think we are fine because we are not heavily configuring the cloudbuild.yaml file. 

As reference: 
```
https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/mvn/examples/spring_boot/cloudbuild.yaml
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,19 +1,8 @@
-# steps:
-# - name: "gcr.io/cloud-builders/mvn"
-#   args:
-#   - package
-# artifacts:
-#   objects:
-#   // This is using sources from teh Google Cloud Source Repository
-#     location: "gs://${_BUCKET}/"
-#     paths:
-#     - "target/${_TARGET}"
-
+# cloudbuild.yaml
 steps:
-- id: 'Staging the app using mvn appengine plugin'
-  name: 'gcr.io/cloud-builders/mvn'
-  args: ['package', 'appengine:stage']
-- id: "Deploy to app engine using gcloud image"
-  name: 'gcr.io/cloud-builders/gcloud'
-  args: ['app', 'deploy', 'codeu-spring-2019-team-9\cloudbuild.yaml']
-timeout: '1600s'
+# Builds the application and deploys it to App Engine using Maven.
+- id: 'build_and_deploy'
+- name: 'gcr.io/cloud-builders/mvn'
+  args: ['package']
+- name: 'gcr.io/cloud-builders/gcloud'
+  args: ['app', 'deploy', 'src/main/webapp/WEB-INF/appengine-web.xml']
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,14 @@
+steps:
+- name: maven:3.6.0-jdk-11-slim
+  entrypoint: 'mvn'
+  args: ['install']
+- name: 'gcr.io/cloud-builders/docker'
+  args:
+    [
+      'build',
+      '-t',
+      'gcr.io/sp19-codeu-9-3673/github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9',
+      '.'
+    ]
+    # This is to save the container images
+images: ['gcr.io/sp19-codeu-9-3673/github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9']
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,14 +1,19 @@
+# steps:
+# - name: "gcr.io/cloud-builders/mvn"
+#   args:
+#   - package
+# artifacts:
+#   objects:
+#   // This is using sources from teh Google Cloud Source Repository
+#     location: "gs://${_BUCKET}/"
+#     paths:
+#     - "target/${_TARGET}"
+
 steps:
-- name: maven:3.6.0-jdk-11-slim
-  entrypoint: 'mvn'
-  args: ['install']
-- name: 'gcr.io/cloud-builders/docker'
-  args:
-    [
-      'build',
-      '-t',
-      'gcr.io/sp19-codeu-9-3673/github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9',
-      '.'
-    ]
-    # This is to save the container images
-images: ['gcr.io/sp19-codeu-9-3673/github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9']
+- id: 'Staging the app using mvn appengine plugin'
+  name: 'gcr.io/cloud-builders/mvn'
+  args: ['package', 'appengine:stage']
+- id: "Deploy to app engine using gcloud image"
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -0,0 +1,14 @@
+steps:
+- name: maven:3.6.0-jdk-11-slim
```
michaelsafyan: Please use the 'gcr.io/cloud-builders/mvn' image.

cloudbuild.yaml
```diff
@@ -0,0 +1,14 @@
...
+  entrypoint: 'mvn'
+  args: ['install']
+- name: 'gcr.io/cloud-builders/docker'
```
michaelsafyan: We don't have a Dockerfile... why is there a 'docker build' step?

What I'd expect, instead, is a 'gcr.io/cloud-builders/gcloud' step to perform a deployment to App Engine (which you might do with Maven, directly, so you probably only need one step).

michaelsafyan (changes requested):
Please note that you can test the `cloudbuild.yaml` file using:
[`gcloud builds submit --config=cloudbuild.yaml`](https://cloud.google.com/sdk/gcloud/reference/builds/submit)

Can you make sure to test it and verify that it works correctly?

cloudbuild.yaml
```diff
@@ -0,0 +1,19 @@
+# steps:
```
michaelsafyan: Can you delete commented-out code?

cloudbuild.yaml
```diff
@@ -0,0 +1,19 @@
...
+# - name: "gcr.io/cloud-builders/mvn"
+#   args:
+#   - package
+# artifacts:
+#   objects:
+#   // This is using sources from teh Google Cloud Source Repository
+#     location: "gs://${_BUCKET}/"
+#     paths:
+#     - "target/${_TARGET}"
+
```
michaelsafyan: Can you add a comment that explains why you do this in multiple steps? That is, why not:

```
steps:
# Builds the application and deploys it to App Engine using Maven.
- id: 'build_and_deploy'
  name: 'gcr.io/cloud-builders/mvn'
  args: ['package', 'appengine:deploy']
  env:
    - "GOOGLE_CLOUD_PROJECT=sp19-codeu-9-3673"
```

I can understand splitting it into multiple steps if, for example, you did some sort of validation step after the build but before the deployment that could be used to prevent the deployment from happening. But since there isn't such a step, it isn't clear to me why you've broken it down into the separate steps like this.

cloudbuild.yaml
```diff
@@ -0,0 +1,19 @@
...
+steps:
+- id: 'Staging the app using mvn appengine plugin'
```
michaelsafyan: Minor note for future reference; the `id` field is intended to be an identifier, not a description.

cloudbuild.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: 'gcr.io/cloud-builders/mvn'
+  args: ['package', 'appengine:stage']
+- id: "Deploy to app engine using gcloud image"
+  name: 'gcr.io/cloud-builders/gcloud'
+  args: ['app', 'deploy', 'codeu-spring-2019-team-9\cloudbuild.yaml']
```
michaelsafyan: Minor note for future reference: the *.yaml file that is expected in a call to `gcloud app deploy` is not the Cloud Build configuration, but rather the configuration for App Engine (that YAML file is automatically generated during the build process from `appengine-web.xml`, `web.xml`, and the other build artifacts).

You can see what such a file looks like in the Python App Engine case here:
https://cloud.google.com/appengine/docs/standard/python/config/appref

The Java App Engine environment uses *.xml in order to be idiomatic (Java uses XML more than YAML), but then that configuration gets converted into YAML as in the Python case.

Based on the documentation for the `appengine:stage` command here:
https://cloud.google.com/appengine/docs/standard/java/tools/maven-reference#appenginestage

... it would appear that the *.yaml file might be output in a subdirectory of `target/appengine-staging`, but it shouldn't be necessary to figure out the details of the mechanics since `appengine:deploy` will do all of this under the hood and magically invoke `gcloud app deploy` with the correct parameters.

markser: > 
> 
> Can you add a comment that explains why you do this in multiple steps? That is, why not:
> 
> ```
> steps:
> # Builds the application and deploys it to App Engine using Maven.
> - id: 'build_and_deploy'
>   name: 'gcr.io/cloud-builders/mvn'
>   args: ['package', 'appengine:deploy']
>   env:
>     - "GOOGLE_CLOUD_PROJECT=sp19-codeu-9-3673"
> ```
> 
> I can understand splitting it into multiple steps if, for example, you did some sort of validation step after the build but before the deployment that could be used to prevent the deployment from happening. But since there isn't such a step, it isn't clear to me why you've broken it down into the separate steps like this.

I tried this approach that was mentioned and wasn't able to figure out get it work as it was producing errors. I understand the logic as that is what we are doing when we are calling ```mvn package appengine:deploy```, but I was unable to get it working with cloudbuild. 

 We might have success if we go with this route:
```
 steps:
# Builds the application and deploys it to App Engine using Maven.
- id: 'build_and_deploy'
   name: 'gcr.io/cloud-builders/mvn'
   args: ['package', 'appengine:deploy']
   env:
     - "GOOGLE_CLOUD_PROJECT=sp19-codeu-9-3673"
   dir: src/main/webapp/WEB-INF/appengine-web.xml
```
This way, we specified which directory to go to. Hopefully this removes the missing directory /workspace/ error when we run the cloudbuild trigger.

markser:
I wasn't able to locally debug/test the cloudbuild.yaml because as of right now you are only able to do so with Linux or macOS. So, I created a new appengine account and tested the cloudbuild.yaml. This was faster/easier than creating a new virtual machine just to test it locally. 

The fix that got the cloudbuild.yaml to work.
I changed the yaml from
```codeu-spring-2019-team-9\cloudbuild.yaml``` -> ```src/main/webapp/WEB-INF/appengine-web.xml```

The one thing with cloud build right now is that it is taking about 1 min and 30 sec avg for it to successfully execute. I will see if I am able to speed up the process by "caching directories with Google Cloud Storage" as seen here ```https://cloud.google.com/cloud-build/docs/speeding-up-builds```. I just wanted to get it working first on our project first before exploring ways to speeding up the cloudbuild.

File: src/main/webapp/WEB-INF/appengine-web.xml [hunk 7]
```diff
@@ -16,6 +16,7 @@
 -->
 
 <appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
+  <application>sp19-codeu-9-3673</application>
```
michaelsafyan, line 19: It should not be necessary to set this tag; the `pom.xml` file specifies how to obtain the application name automatically.

michaelsafyan (commented):
Can you describe for me how you tested/verified this?

PR #17, opened 2019-03-19
Finished creating charts using JSON data

markser:
Finished the guided steps on Charts 2.

- Clean up the code in chart-loader.js
- Need to figure out how to created a form for users to use. Most likely going to use html
- How to have distinction with user stats and global stats
- Combine the stats and charts page
- Remove the .html


michaelsafyan:
I will wait on @paulinagzd to review before commenting.

paulinagzd (approved):
Everything looks okay, I liked the intuitive way of displaying messages. Well commented and the rest is self-explanatory.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -100,12 +100,20 @@ public void storeMessage(Message message) {
     // people may include implicit context/reference to previously sent messages,
     // and this order makes the chronology of the conversation and context
     // more intuitive and understandable in the messages list page. 
-    Query query =
+    Query query;
```
michaelsafyan: I see two issues here:

   1. Both branches of the `if` are identical. Therefore, the branch is not needed.
   2. Even if conditional logic were necessary, a declaration without an assignment statement is generally a code smell; when conditional initialization logic is necessary, use a subroutine and/or ternary to always put an assignment in the declaration statement. As in:

```
Query query = hasUserRestriction(loggedInUser, otherUser)
                        ? createMessageQueryWithUserFilter(loggedinUser, otherUser)
                        : createPublicMessageQuery();
```

src/main/java/com/google/codeu/servlets/ChartServlet.java
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Responds with a JSON representation of {@link Message} data for logged-in user and another user. Responds with
+   * an empty array if the user is not logged in or another user is not provided.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    response.setContentType("application/json");
+
+    List<Message> messages = datastore.getMessagesBetweenTwoUsers(null, null);
```
michaelsafyan: Interfaces should generally disallow null. This usage also makes the implementation of this operation needlessly complex and confusing (it is especially weird that "messages between two users" would allow either user to be omitted). Instead, create a different `getAllMessages` operation for your intended use case, please.

src/main/java/com/google/codeu/servlets/ChartServlet.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /*
+   * Instantiates a connection to a datastore. In tests, override this with a
+   * method that returns a fake/mock datastore object.
+   */
+  protected Datastore createDatastore() {
+    return new Datastore();
+  }
+
+  /**
```
michaelsafyan, line 52: How is this different than the existing messages servlet?

src/main/java/com/google/codeu/servlets/ChartServlet.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.List;
+
+import javax.servlet.ServletConfig;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Whitelist;
+
```
michaelsafyan, line 33: Add a comment explaining what this servlet is intended to do?

src/main/webapp/WEB-INF/web.xml
```diff
@@ -65,4 +70,10 @@
     <url-pattern>/stats</url-pattern>
     <servlet-name>stats</servlet-name>
   </servlet-mapping>
+
+  <servlet-mapping>
+    <url-pattern>/messagechart</url-pattern>
```
michaelsafyan: I'd recommend `/api/chart` (with `/chart` being the URL for the corresponding *.html code).

src/main/webapp/js/chart-loader.js
```diff
@@ -79,13 +82,27 @@
...
 
+  function fetchMessageData() {
+    const url = "/messagechart";
+    if(_messageChartPromise == null) {
+      _messageChartPromise = fetch(url)
+      .then((response) => { 
+        return response.json();
+      })
+      .then((msgJson) => {
+        console.log(msgJson);
```
michaelsafyan: Also add:

     return msgJson

So that the object is passed along to whatever function is registered with calling `.then` on the result of `fetchMessageData()`.

src/main/java/com/google/codeu/data/Datastore.java
markser: I did not know that someone's code could even smell. TIL

src/main/java/com/google/codeu/servlets/ChartServlet.java
markser: The Chart Servlet passes the data as a response to a GET request.

src/main/java/com/google/codeu/servlets/ChartServlet.java
markser: Created a new getAllMessages function to resolve this.
The original intent was that it would be a good idea to reuse the function that was already created, but I see the weirdness of a function accepting nulls and outputting something.

src/main/java/com/google/codeu/servlets/ChartServlet.java
markser: Now that I look at it, there isn't a huge difference between the message and the chart servlet. Considering that I am handling message data on the chart servlet, I will see what I can do to make the two servlets more abstract from each other.

src/main/webapp/js/chart-loader.js
```diff
@@ -22,19 +22,16 @@
             //TODO add the formatted values to msgRow array by using JS' push method
             msgData.addRow(msgRow);
         }
-        drawChart(msgData);
+        var chart = new google.visualization.BarChart(
```
markser: https://github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9/pull/17#commitcomment-32878337

Hopefully fixed the race condition

markser:
https://github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9/pull/17#commitcomment-32878337
- I created a new ``` private List<Message> entityToMessage(List<Message> messages, PreparedQuery results) { ``` and the necessary methods to make sure that the ternary operator will work.

https://github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9/pull/17#commitcomment-32878373

- I believe that I fixed the race condition in ```chart-loader.js```. I removed ```drawChart()``` and implemented the functionality into the promise ```.then(msgJson)``` of ```fetchMessageData```

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -100,12 +103,33 @@ public void storeMessage(Message message) {
...
+    
+    Query query = 
+        new Query("Message")
+          .addSort("timestamp", SortDirection.ASCENDING);
+    
+    PreparedQuery results = datastore.prepare(query);
+    
+    return entityToMessage(messages, results);
+  }
+  private List<Message> entityToMessage(List<Message> messages, PreparedQuery results) {
```
michaelsafyan: Minor feedback: `entityToMessage` implies a single `Entity` is being converted to a single `Message`.  A method like this should probably be named `preparedQueryToMessages`. It is, in my view, confusing that data is returned through an output parameter. If you have an output parameter, it would be better to return `void`. However, even better would be to return data through the `return` statement only, and not via output parameters.

My recommendation to you:

```
  private List<Message> preparedQueryToMessages(PreparedQuery query) {
     List<Message> results  = new ArrayList<>();
     for (Entity entity : query.asIterable()) {
        results.add(entityToMessage(entity));
     }
     return results;
  }

  private Message entityToMessage(Entity entity) {
      String idString = entity.getKey().getName();
      UUID id = UUID.fromString(idString);
      String text = getStringProperty(entity, "text").orElse("");
      long timestamp = (long) entity.getProperty("timestamp");
      String sender = (String) entity.getProperty("user");
      String receiver = (String) entity.getProperty("recipient");
      return new Message(id, sender, text, timestamp, receiver);
  }
```

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -116,15 +140,23 @@ public void storeMessage(Message message) {
...
-  /** Returns the total number of messages for all users. */
-  public int getTotalMessageCount() {
-    Query query = new Query("Message");
-    PreparedQuery results = datastore.prepare(query);
-    return results.countEntities(FetchOptions.Builder.withDefaults());
-  }
+ }
+
+ private boolean hasUserRestriction(String loggedInUser, String otherUser) {
+   if (!Strings.isNullOrEmpty(loggedInUser) && !Strings.isNullOrEmpty(otherUser)) {
```
michaelsafyan: The following pattern:

```
      if (expr) {
        return false;
      }
      return true;
```

Can be reduced to:

```
    return !expr;
```

That being said, it looks like you've reversed `false` and `true` (so it should be reduced to `return expr`).

Simplify to:?

```
private boolean hasUserRestriction(String loggedInUser, String otherUser) {
   return !Strings.isNullOrEmpty(loggedInUser) && !Strings.isNullOrempty(otherUser);
}
```

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -100,12 +103,33 @@ public void storeMessage(Message message) {
...
+    List<Message> messages = new ArrayList<>();
+    
+    Query query = 
+        new Query("Message")
+          .addSort("timestamp", SortDirection.ASCENDING);
+    
+    PreparedQuery results = datastore.prepare(query);
+    
+    return entityToMessage(messages, results);
+  }
```
michaelsafyan: Please leave a blank line before each function.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -116,15 +140,23 @@ public void storeMessage(Message message) {
...
-  }
+ }
+
+ private boolean hasUserRestriction(String loggedInUser, String otherUser) {
+   if (!Strings.isNullOrEmpty(loggedInUser) && !Strings.isNullOrEmpty(otherUser)) {
+     return false;
+   }
+   return true;
+ }
+ private Query createMessageQueryWithUserFilter(String loggedInUser, String otherUser, Filter directMessages) {
```
michaelsafyan: This function does not use the `loggedInUser` or `otherUser` parameters. Why are they passed-in?

My recommendation would be:

```
   private Query createMessageQueryWithUserFilter(String loggedInUser, String otherUser) {
      return Query("Message")
         .setFilter(createUserFilter(loggedInUser, otherUser))
         .addSort("timestamp", SortDirection.ASCENDING);
   }
```

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -100,12 +103,33 @@ public void storeMessage(Message message) {
     // people may include implicit context/reference to previously sent messages,
     // and this order makes the chronology of the conversation and context
     // more intuitive and understandable in the messages list page. 
-    Query query =
-        new Query("Message")
-            .setFilter(directMessages)
-            .addSort("timestamp", SortDirection.ASCENDING);
+    Query query = hasUserRestriction(loggedInUser, otherUser)
```
michaelsafyan: Now that there is a `getAllMessages()` function, is there a need to have this conditional logic?

Instead, I would recommend inserting:

```
   Preconditions.checkArgument(!Strings.isNullOrEmpty(loggedInUser), "loggedInUser is null or empty");
   Preconditions.checkArgument(!Strings.isNullOrEmpty(otherUser), "otherUser is null or empty");
```

... at the very top of this function and not accepting this null/empty case.

src/main/webapp/js/chart-loader.js
```diff
@@ -1,4 +1,37 @@
...
+          msgData.addColumn('number', 'Message Count');
+          for (i = 0; i < msgJson.length; i++) {
+            msgRow = [];
+            var timestampAsDate = new Date(msgJson[i].timestamp);
+            var totalMessages = i + 1;
+            msgRow.push(timestampAsDate, totalMessages);
+            //TODO add the formatted values to msgRow array by using JS' push method
+            msgData.addRow(msgRow);
+          }
+          var chart = new google.visualization.BarChart(
```
michaelsafyan: Can you move the call to DOM operations  (like `getElementById`) out of `fetchMessageData`?

Instead, there should be code like:

    `var drawChart = function() {
          fetchMessageData()
              .then(function (msgData) {
                  // ... do DOM calls like `getElementById` and `draw`
              });
       };

And then you can invoke `drawChart` in the `onload` code, while calling `fetchMessageData` earlier.

src/main/webapp/js/chart-loader.js
```diff
@@ -79,18 +112,17 @@
...
   };
 
-
   var init = function() {
     google.charts.load("current", {
       packages: ["corechart"],
     });
     google.charts.setOnLoadCallback(
       function() {
+        fetchMessageData();
```
michaelsafyan: Retrieving the message can happen sooner. You only need to wait for the Google Chart's onload event for using the types defined by that API. You should be able to initiate the backend `fetch` directly within `init`.

michaelsafyan (changes requested):
It looks like you have a bunch of unrelated formatting changes here... please make a formatting-only change in a separate pull request, first, before making this change so that the "diff" only shows substantial changes and isn't full of noise resulting from format-only changes.

src/main/java/com/google/codeu/data/Datastore.java
markser: Yeah, this makes more sense than the spaghetti code I wrote.
Thank you for the feedback!

src/main/java/com/google/codeu/data/Datastore.java
markser: This makes sense and makes the code cleaner. Perfect!

src/main/java/com/google/codeu/data/Datastore.java
markser: Created a new function ```createUserFilter(loggedInUser, otherUser))```
Thank you for your recommendation!!!

src/main/java/com/google/codeu/data/Datastore.java
markser: I did not know that ```Preconditions.checkArgument()``` was a thing. I will take a look and implement it.
Thank you for letting me know

src/main/webapp/js/chart-loader.js
markser: Okay, I just want to see if I am understanding this right now after reading up on everything again and combing through the past comments as well.

So, when chart-loader.js is running, it’s going to first run everything that is in ```init()``` first and this is where want to call ```fetchMessageData()``` so that we fetch our data first. Then, with the help of ```.setOnLoadCallback()``` , it will call ```drawChart()``` after we finish calling ```init()```. This will prevent the potential run condition that we are going to have because we are getting the chart data before drawing the chart. Not like before, where we are drawing the chart and fetching the message data at the same time because the code we had before was asynchronous.  

If this is right, then all of this makes more sense now.

markser:
I now merged the userChart ()html form for the user) and charts2 together

src/main/java/com/google/codeu/data/Tea.java
```diff
@@ -0,0 +1,84 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.codeu.data;
+
+// import java.util.UUID;
+
+/** A single message posted by a user. */
+public class Tea {
```
michaelsafyan: This solution seems overly complicated. How about:

```
public enum TeaCategory {
   GREEN,
   WHITE,
   BLACK,
   HERBAL;
}
```

And then use `Map<TeaCategory, Integer>` to represent a histogram of each category rather than maintaining separate integer variables for each and every category?

src/main/java/com/google/codeu/data/Tea.java
```diff
@@ -0,0 +1,84 @@
...
+    
+    private String userName;
+    private int greenTea;
+    private int whiteTea;
+    private int yellowTea;
+    private int oolongTea;
```
michaelsafyan: Aren't oolong and matcha types of green tea? It sounds like you may need a tree-like structure or hierarchy if you are going to count specific subtypes of tea.

src/main/java/com/google/codeu/servlets/UserFormServlet.java
```diff
@@ -0,0 +1,118 @@
...
+  protected Datastore createDatastore() {
+    return new Datastore();
+  }
+
+@Override
+public void doPost(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
+
+  UserService userService = UserServiceFactory.getUserService();
+    if (!userService.isUserLoggedIn()) {
+      response.sendRedirect("/index.html");
```
michaelsafyan: This is a soft failure. It is better to do a hard failure by emitting a `401` status code (which means "unauthenticated").

To make this a good user experience with a hard failure, the `POST` should be done in JavaScript (where the JavaScript code can handle the failure), rather than having this URL be hit directly by the browser as part of the form's action. However, although it means a little more client-side work for the web case, it is still worth using a hard failure rather than a soft failure here, because it then makes client-side detection of the error much more reliable for other kinds of clients (e.g. if you were to create a mobile application or a script that hit this endpoint, redirects are typically followed automatically and are easily miscategorized as "success", while setting an appropriate error status code will make it easy for code that fetches the page programatically to realize that something has gone wrong).

src/main/java/com/google/codeu/servlets/UserFormServlet.java
```diff
@@ -0,0 +1,118 @@
...
+      return;
+    }
+    String user = userService.getCurrentUser().getEmail();
+    int greenTea = Integer.parseInt(request.getParameter("greenTea"));
```
michaelsafyan: What happens if the parsing fails? It looks like this will result in a `500` (internal server error) status code being generated due to an uncaught exception. However, the appropriate response should be a `400` (bad request).

I would suggest doing this parsing with some error-handling logic that will emit an appropriate `400` status code (with content that indicates the part of the request that failed to validate correctly).

src/main/java/com/google/codeu/servlets/UserFormServlet.java
```diff
@@ -0,0 +1,118 @@
...
+    Tea userTeaData = new Tea(user,greenTea,whiteTea,yellowTea,oolongTea,blackTea,matchaTea);
+
+    datastore.storeUserFormData(userTeaData);
+  }
+}
+  
+
+
+
+  // Entity post = new Entity("userFormData"); // create a new entity
```
michaelsafyan: Delete commented code, please.

src/main/webapp/user-page.html
```diff
@@ -16,29 +16,67 @@
...
+        <li><a href="/">Home</a></li>
+      </ul>
+    </nav>
+    <h1 id="page-title">User Page</h1>
+   
+    <jsp:directive.taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
+    <jsp.directive.taglib uri="http://java.sun.com/jsp/jstl/functions" prefix="fn"%>
+<div class="userForm">
+  <h2>
+    Input what tea you drank today
```
michaelsafyan: Minor wording suggestion: I think it feels a little more natural and less demanding to phrase it as a question, like "What tea did you drink today?"

src/main/webapp/user-page.html
```diff
@@ -16,29 +16,67 @@
...
+  </h2>
+
+  <form method="POST" action="/api/user-form-data">
```
michaelsafyan: I would remove the `method` and `action` properties and, instead, handle form submission in JavaScript. To handle in JavaScript, change:

```
<button type="submit">Save</button>
```

To:

```
<button onclick='onSubmit'>Save</button>
```

And then define a function called `onSubmit` in your JavaScript code.

An alternative approach is to create a CSS selector on the button and then register the action in JavaScript:

```
// JavaScript
var submitButton = document.querySelector('#submit-button');
submitButton.addEventListener('click', onSubmit);

<!-- HTML -->
<button id="submit-button">Save</button>
```

src/main/webapp/user-page.html
```diff
@@ -16,29 +16,67 @@
...
+
+    <div>
+      <label for="greenTea">Green Tea</label>
+      <input type="number" name="greenTea" id="greenTea" size="40" value="${fn:escapeXml(tea.greenTea)}" class="form-control" />
```
michaelsafyan: What's this escaping for?

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -161,4 +161,21 @@ public int getTotalMessageCount() {
     return results.countEntities(FetchOptions.Builder.withDefaults());
   }
 
+  /**
+   * This is to create an user form data to put in the datastore
+   * TODO: Replace getUser() with UUID and use userChart.getId().toString()
+   */
+  public void storeUserFormData(Tea userChart) {
+    
+    Entity userFormData = new Entity("UserFormData", userChart.getUserName());
```
michaelsafyan: This entity name seems off to me. I would expect it more to be like `UserTeaConsumption`, `TeaConsumptionByUser`, or `TeaConsumption`. That there is no mention of tea at all is surprising.

michaelsafyan: Given that the form asks about "today", not "all time", I'd expect that this entity would have a time component to it. I'd recommend either incorporating the date into the key along with the user, or I'd recommend making both 'user' and 'date' into properties of the entity, relying on the datastore to auto-assign a key (this, though, would mean that you would need to do a query to update the entity for a given user on a given date).

michaelsafyan (changes requested):

src/main/webapp/user-page.html
markser: So, one user on stackoverflow mention that one of the advantages of using a submit button was if a user doesn't have JavaScript enabled in their browser, the form would still work. This is why right now I have it where JavaScript is not handling the form submission yet. 

Should I just assume that in our scope, that all potential users will have JavaScript enabled within their browsers and that they wouldn't run into this issue?

https://stackoverflow.com/questions/16050798/html-form-submit-using-javascript-vs-submit-button

src/main/webapp/user-page.html
markser: We are using JavaServer Page's ```escapeXml``` to counter XSS attacks. 

It's explained here : https://cloud.google.com/appengine/docs/standard/java/building-app/handling-form-data

``` 
This form sends the form data to the handler at the /create URL. 
You should use JSP's escapeXml functionality to counter cross-site scripting (XSS) attacks
 ```

File: src/main/webapp/user-page.html (hunk 21 above)
Comment (michaelsafyan) [reply src/main/webapp/user-page.html]: Yes, this is accurate and known as "graceful degradation"; web apps that need to target browsers with and without JavaScript will often have a handler in JavaScript (which can provide more immediate validation and feedback without issuing a network request) as well as an action that points to a server-side submit form that is invoked in the JavaScript-less case.

That being said, JavaScript is supported in all modern browsers, and most modern websites assume that JavaScript is supported (and often will not work without JavaScript). Given that most of the other code already implemented has a dependency on JavaScript with no graceful degradation (e.g. the login functionality is implemented in a such a way that JavaScript is required), I would not spend a whole lot of time implementing graceful degradation.

In a real application, it is typical to first target the vast majority of cases (assume a modern browser, assume JavaScript is supported, etc., etc.) and only then, later, after getting this minimum viable product working, look to extend support to older browsers or browsers that have JavaScript disabled.

File: src/main/webapp/user-page.html (hunk 22 above)
Comment (michaelsafyan) [reply src/main/webapp/user-page.html]: Thanks for the link.  So, I think there are two things that you are missing:

  1. In the example, the assumption is that the HTML code is being generated on the server
      from a template using JSP. That is, the example works because "${fn:...}" is being
      evaluated in some code that is processing the `*.jsp` file before returning it as an
      HTML file to the browser. In our code, there is no such preprocessing; the *.html file
      is being returned as a static file with no such dynamic, server-side evaluation (unless you've
      specially wired it up like that? If so, please point me to where in your code, you are
      causing the HTML to be processed using JSP). 

  2. The attack is possible and needs to be mitigated, because arbitrary user-supplied
      text was being injected into the HTML response, and this text could potentially
      include content like `<script>` tags that could execute arbitrary code. In this case,
      though, you are including numeric values that cannot possibly include `<script>` tags.

File: src/main/webapp/user-page.html (hunk 21 above)
Comment (markser) [reply src/main/webapp/user-page.html]: Got it, it would be a headache to implement a fallback for the users if we already created code that already assumes that the user will have JavaScript enabled. 

Thank you for the comment! It's interesting to read about the difference between "graceful degradation" and "progressive enhancement".

File: src/main/webapp/user-page.html (hunk 22 above)
Comment (markser) [reply src/main/webapp/user-page.html]: If I am understanding this correctly for **2**; If we to call for the form to only accept number values, then we don't need ```escapeXml``` because it's not possibly to include ```<script>``` tags in that field. Preventing the XSS attack from happening in the user form. 

However, if I was to change the input type to text for **1**, then we would have to use the escapeXml functionality to counter the XSS attack. On top of that, I would need to implement something like server side rendering onto ```user-page.html``` so that we are processing the input (*I did not implement server side rendering / preprocessing*).

If it's the case where I don't have to worry about it in this case if we keep our input type as numbers, I think I can safely remove it.

Thank you for the insight and the comment!

File: src/main/java/com/google/codeu/data/Datastore.java (hunk 23 above)
Comment (markser) [reply src/main/java/com/google/codeu/data/Datastore.java]: Fixed the entity name from  ```userFormData``` -> ```userTeaConsumption```

I agree, it was weird for the name to not mention anything about the tea as it is taking/holding the user's tea data.

File: src/main/java/com/google/codeu/data/Tea.java (hunk 16 above)
Comment (markser) [reply src/main/java/com/google/codeu/data/Tea.java]: For now, I am just going to have Green, White, Black, and Herbal tea. 
I will go back to  implementing the children for the sub teas when I finish the framework for the discussion / threads for the website

Comment (michaelsafyan) [reply src/main/java/com/google/codeu/data/Tea.java]: That sounds like a good idea. Simpler is better.

File: src/main/java/com/google/codeu/data/Datastore.java [hunk 24]
```diff
@@ -184,20 +184,17 @@ public void storeUserTeaData(
       datastoreUser.setProperty("username", datastoreUser.getProperty("username"));
       datastoreUser.setProperty("date", datastoreUser.getProperty("date"));
 
-      Map<String, Long> newMap = new HashMap<String, Long>();
       EmbeddedEntity teaMap = (EmbeddedEntity) datastoreUser.getProperty("teaData");
```
markser: This is valid work around for not able to directly using a Map, but I should look into this when I have the time.
https://github.com/objectify/objectify/wiki/Entities

markser:
I still need to work on 
- https://github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9/pull/17#discussion_r270426355
- https://github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9/pull/17#discussion_r270427294
- https://github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9/pull/17#discussion_r270430106

But the main idea of storing the userTeaData into the datastore with the key being the date+username is all done. 
Need to finish form handling and parsing with creating a chart with the user data now. 

I should've probably just have the Tea data into a JSON, but what's coding if you do the easy thing first and then rebuild everything :D :(  -_-

src/main/java/com/google/codeu/servlets/UserFormServlet.java
```diff
@@ -60,25 +63,43 @@ protected Datastore createDatastore() {
...
 
-    UserService userService = UserServiceFactory.getUserService();
-    if (!userService.isUserLoggedIn()) {
-      response.setStatus(401);
-      response.getWriter().println("Error: Unauthorized access");
-      return;
-    }
-    String username = userService.getCurrentUser().getEmail();
+        UserService userService = UserServiceFactory.getUserService();
+        if (!userService.isUserLoggedIn()) {
```
markser, line 72: I believe that this is a temp solution to https://github.com/codeu-spring-2019-team-9/codeu-spring-2019-team-9/pull/17#discussion_r270426355
But I am still stuck on trying to figure out how to do a hard failure with a 401 status code on the client side,
I was thinking of using a try and catch block and when it catches that the user is not logged in, it would then emitt error 401. But I am currently stuck on seeing how I can do that in JavaScript if that is the correct way of doing this.

michaelsafyan: Please see the documentation for the JavaScript `fetch` function:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

As well as for the `Response` object that is returned as the result of the `Promise`:
https://developer.mozilla.org/en-US/docs/Web/API/Response

And this documentation for the Promise API:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise
https://developers.google.com/web/fundamentals/primers/promises

In particular, from the documentation for `fetch`, note:

> The `fetch()` method takes one mandatory argument, the path to the resource you want to fetch. It returns a Promise **that resolves to the `Response` to that request, _whether it is successful or not_**. 

And, from the documentation for `Response`, please note the properties `ok`, `status`, and `statusText`. Here is a snippet of my own that should help:

```
     var handleSuccessfulResponse = function(response) {
        console.log('Reponse was successfull');
        // ... do stuff with valid response ...
     };

     var handleErrorResponse = function(response) {
        console.log('Response was unsuccessful'); 
        console.log('Status code: ' + response.status);
        console.log('Status message: ' + response.statusText);
        // ... do stuff with error response ...
     };

     // A version of "fetch" that returns a Promise that splits
     // responses so that "then" only receives successful responses
     // and "catch" only receives unsuccessful responses.
     var fetchWithErrorHandling = function(url) {
       return fetch(url).then(function(response) {
         if (!response.ok) {
           console.log('Response is not OK');
           throw response;
         }
         console.log('Response is OK');
         return response;
       });
     };
       
    // ...
    var eventualResponse = fetchWithErrorHandling(someUrl);
    eventualResponse.then(handleSuccessfulResponse);
    eventualResponse.catch(handleErrorResponse);
    // ...   
```

src/main/java/com/google/codeu/servlets/UserFormServlet.java
```diff
@@ -81,15 +81,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
         for (String tea : teaNames) {
           try {
             String formTea = request.getParameter(tea);
-            if (formTea == null || !formTea.isEmpty()) {
-              throw new NumberFormatException();
+            if (Strings.isNullOrEmpty(formTea)) {
```
markser: I am parsing through the parameters and checking if they are null or empty using Guava's null or empty function and throwing a exception if it's true. I am getting an ```500`` error that is caused by a NullPointerException 

The errors are happening here:
Datastore.java:192
UserFormServlet.java:100

Am I going to have to put a stop on trying to fix this bug and work on getting the form validation to on client side and come back to trying to fix this bug afterwards? As of right now, it seems to me that this is probably what I have to do.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,130 @@ public void storeMessage(Message message) {
...
+   *
+   * @throws EntityNotFoundException
+   */
+  public void storeUserTeaData(
+     Map<String, Long> incomingUserTeaData, String username, String date) {
+
+    Entity datastoreUser;
+    Key usernameKey = KeyFactory.createKey("UserTeaData", date + username);
+
+    try {
```
markser: This is parsing through the datastore to check to see if if the user has already filled out a form for the day and will update the user's amount of tea for the day.
If it can't find the user's tea data for the day, it will create a new entity with the user's tea data.

The only thing that I think is flawed is that I am just setting the sent in ```teaMap``` directly instead of checking it, but I think it's fine if I have ```UserFormServlet``` make sure that it will at least send in a Map initialized with the keys and values with 0 so that we don't have an updating failure here.

src/main/webapp/js/user-page-loader.js
```diff
@@ -34,37 +34,37 @@ function setPageTitle() {
  */
```
markser, line 35: I was wondering, should we have the ```user-page-loader.js``` follow this format as well?
```
(function(){
   // ... code here ...
})();
```
I guess my question is that should we always have our JavaScript files look like this because it prevents the possibility of having variables global? 
Side note, shouldn't we have it where we fetch our messages first then to build the message div? It seems to me, as of right now, we have a race condition with the messages.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
 
-    // Messages are ordered from oldest to newest since messages between
-    // people may include implicit context/reference to previously sent messages,
-    // and this order makes the chronology of the conversation and context
-    // more intuitive and understandable in the messages list page. 
-    Query query =
-        new Query("Message")
-            .setFilter(directMessages)
-            .addSort("timestamp", SortDirection.ASCENDING);
+  public List<Message> getAllMessages() {
```
michaelsafyan: Minor comment: can you put `public` methods before `private` methods? This way, the file is more scannable in terms of which methods are available to other files.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
-    for (Entity entity : results.asIterable()) {
-      String idString = entity.getKey().getName();
-      UUID id = UUID.fromString(idString);
-      String text = getStringProperty(entity, "text").orElse("");
-      long timestamp = (long) entity.getProperty("timestamp");
-      String sender = (String) entity.getProperty("user");
-      String receiver = (String) entity.getProperty("recipient");
-      Message message = new Message(id, sender, text, timestamp, receiver);
-      messages.add(message);
+  private List<Message> prepareQueryToMesssages(PreparedQuery query) {
```
michaelsafyan, line 199: Minor naming comment: "prepare" -> "prepared". Otherwise, it seems that "prepare" is the verb/action being taken by this function (rather than "preparedQuery" being a single composite noun, referring to the input).

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
   }
 
+  /**
+   * This is to create an user form data to put in the datastore Going with option 2 highlighted in
+   * the Meeting Notes
+   *
+   * @throws EntityNotFoundException
+   */
+  public void storeUserTeaData(
+     Map<String, Long> incomingUserTeaData, String username, String date) {
```
michaelsafyan: Using raw strings for times is an easy way to make mistakes. I would recommend using more structured representations for time in all of the code except for I/O where the information, being exchanged with the world outside of your code, necessarily needs to be converted to/from unstructured strings.

I'd suggest using [`java.time.LocalDate`](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html) `date` and [`java.time.ZoneId`](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html) `timeZone` parameters.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
+
+    Entity datastoreUser;
+    String keyName = date + ":" + username;
+    Key usernameKey = KeyFactory.createKey("UserTeaData", keyName);
```
michaelsafyan: Names of entity kinds and entity properties should be stored in constants. Ex:

     // DatastoreConstants.java
     public class DatastoreConstants {
        public class UserTeaData  {
          public static final String KIND = "UserTeaData";
          public static final String USERNAME_PROPERTY = "username";
          public static final String DATE_PROPERTY = "date";
       }
       // ...
     }

It's also fine to put these constants directly in this file, but they should be class-level `static final` fields.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
+  /**
+   * This is to create an user form data to put in the datastore Going with option 2 highlighted in
+   * the Meeting Notes
+   *
+   * @throws EntityNotFoundException
+   */
+  public void storeUserTeaData(
+     Map<String, Long> incomingUserTeaData, String username, String date) {
+
+    Entity datastoreUser;
```
michaelsafyan: The `datastoreUser` variable name is misleading; the variable does not represent "a user of the datastore" or "a user stored in the datastore", but rather the "tea data for a given user on a given date".  Since there are no other `Entity` variables in the local scope, I don't think it is necessary to disambiguate, and I would simply call this variable `entity`.

michaelsafyan: Please always put an `=` sign when declaring a variable. `Entity entity = null;` not simply `Entity entity;`.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
+    String keyName = date + ":" + username;
+    Key usernameKey = KeyFactory.createKey("UserTeaData", keyName);
+
+    try {
+      datastoreUser = datastore.get(usernameKey);
```
michaelsafyan: When it comes to `try`...`catch`, it is best to minimize the amount of content within the `try` block. When there are too many statements in a single `try` block, it increases the number of ways that the try might fail resulting in an exception being caught in the corresponding handler code, and it makes it confusing to other readers regarding which of the various possible sources of error you were intending to handle.

I would recommend simplifying it like this:

    public void storeUserTeaData(
        String username,
        LocalDate date,
        ZoneId timeZone,
        Map<String, Integer> histogram) {
       Key key = createTeaUserDataKey(username, date, timeZone);
       Entity entity = null;
       try {
         entity = datastore.get(key);
       } catch (EntityNotFoundException e) {
         entity = createNewUserTeaDataEntity(key, username, date, timeZone);
       }
       setUserTeaDataHistogram(histogram, entity);
       datastore.put(entity);
    }

That is, your `try` should only wrap the `get` operation. Additionally, by creating subroutines here, you can minimize repetition; for example, the code to populate the histogram into the entity is shared on both code paths and should not need to be repeated.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
+      datastoreUser.setProperty("username", datastoreUser.getProperty("username"));
```
michaelsafyan: When the `.get()` operation succeeds, it should not be necessary to re-write the `username` and `date` properies, as these cannot possibly change on an update (as they are part of the identity of the entry).

michaelsafyan: When the `.get()` operation succeeds, it should not be necessary to re-write the `username` and `date` properies, as these cannot possibly change on an update (as they are part of the identity of the entry).

michaelsafyan: When the `.get()` operation succeeds, it should not be necessary to re-write the `username` and `date` properies, as these cannot possibly change on an update (as they are part of the identity of the entry). Also, calling `set` with the result of `get` is effectively a no-op.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
+      }
+
+      datastoreUser.setProperty("teaData", teaMap);
+      datastore.put(datastoreUser);
+
+    } catch (EntityNotFoundException e) {
+
+      Entity userTeaConsumption = new Entity(usernameKey);
+      userTeaConsumption.setProperty("username", username);
+      userTeaConsumption.setProperty("date", date);
```
michaelsafyan: Note that `Entity` directly supports properties of type `java.util.Date` for timestamps. It is ideal to use that type to represent the property in the entity, because that will enable indexing/querying by date in a semantic way.

michaelsafyan: Note that `Entity` directly supports properties of type `java.util.Date` for timestamps. It is ideal to use that type to represent the property in the entity, because that will enable indexing/querying by date in a semantic way.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -73,58 +83,131 @@ public void storeMessage(Message message) {
...
+    try {
+      datastoreUser = datastore.get(usernameKey);
+      datastoreUser.setProperty("username", datastoreUser.getProperty("username"));
+      datastoreUser.setProperty("date", datastoreUser.getProperty("date"));
+
+      EmbeddedEntity teaMap = (EmbeddedEntity) datastoreUser.getProperty("teaData");
+
+      for (String key : teaMap.getProperties().keySet()) {
+        Long value1 = incomingUserTeaData.get(key);
+        Long value2 = value1 + (Long) teaMap.getProperty(key);
```
michaelsafyan: Why are you assuming that it is additive? That doesn't seem to reflect the nature of the form. I would assume that the data should simply be overwritten with the new values.

michaelsafyan: Why are you assuming that it is additive? That doesn't seem to reflect the nature of the form. I would assume that the data should simply be overwritten with the new values.

michaelsafyan: Why are you assuming that it is additive? That doesn't seem to reflect the nature of the form. I would assume that the data should simply be overwritten with the new values.

michaelsafyan: Why are you assuming that it is additive? That doesn't seem to reflect the nature of the form. I would assume that the data should simply be overwritten with the new values.

michaelsafyan: Why are you assuming that it is additive? That doesn't seem to reflect the nature of the form. I would assume that the data should simply be overwritten with the new values.

src/main/java/com/google/codeu/servlets/ChartServlet.java
michaelsafyan: This was marked as resolved, but it looks like this file is still here?

src/main/java/com/google/codeu/servlets/UserFormServlet.java
michaelsafyan: Rather than throw an exception, just set the response status and return from `doGet`.

src/main/webapp/user-page.html
```diff
@@ -16,31 +16,60 @@
...
-      <textarea name="text" id="message-input"></textarea>
-      <br/>
-      <input type="submit" value="Submit">
+    </ul>
+</nav>
+<h1 id="page-title">User Page</h1>
+
+<div class="userForm">
+    <h2>
+        What tea did you drink today?
```
michaelsafyan: Given that the interpretation of "today" is dependent on the user's timezone, this form needs to have some way to propagate the date and time-zone of the user back to the server. If submitting this form in JavaScript, you could do this trivially (by retrieving that information and passing it along in the request). If you are using the built-in form action, and `<button type="submit">`, then you might be able to accomplish this by injecting an `<input>` with `type="hidden"` , where you have precomputed the date/timezone information and stored it in advance of the user hitting the submit button.

michaelsafyan (changes requested):
Thanks. This looks much better than before and is clearly on the right track. A few more comments.

src/main/java/com/google/codeu/data/Datastore.java
markser: I see your point and I think that it would probably be a better idea to update the values in the form so that the user knows what their current count of tea is for the day

I was initially thinking that the user might want to update their values throughout the day. So in the morning, they put in that they had ```[greenTea:5, blackTea:0, whiteTea:2, herbalTea:0]```. At the end of the day, they had more tea and wanted to log it in. If we had it where it would outright replace the values, then we could end up with a situation with this ```[greenTea:0, blackTea:0, whiteTea:0, herbalTea:5]``` instead of  ```[greenTea:5, blackTea:0, whiteTea:2, herbalTea:5]```. This is why I had it additive

src/main/java/com/google/codeu/data/Datastore.java
markser: They do not work, the entity will not allow them to be properties. I have just converted ```java.time.ZonedDateTime``` into ```import java.util.Date``` and that works. ```java.time.ZonedDateTime``` has all of the information that we will need, including the timezone of the user. 

```Date date = Date.from(java.time.ZonedDateTime.now().toInstant());```
date stores: ```Mon Apr 22 04:45:54 UTC 2019```

PR #18, opened 2019-03-24
Sentiment analysis

shamailah:
Added a sentiment score for each message sent between users

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,7 @@ limitations under the License.
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-language</artifactId>
-      <version>1.55.0</version>
+      <version>1.64.0</version>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -67,10 +67,10 @@ limitations under the License.
     </dependency>
 
     <dependency>
-        <groupId>com.google.flogger</groupId>
-        <artifactId>flogger-system-backend</artifactId>
-        <version>0.3.1</version>
-        <scope>runtime</scope>
+      <groupId>com.google.flogger</groupId>
+      <artifactId>flogger-system-backend</artifactId>
+      <version>0.3.1</version>
+      <scope>runtime</scope>
     </dependency>
 
     <dependency>
--- a/src/main/java/com/google/codeu/data/Message.java
+++ b/src/main/java/com/google/codeu/data/Message.java
@@ -26,21 +26,23 @@ public class Message {
   private String text;
   private long timestamp;
   private String recipient;
+  private float sentimentScore;
 
   /**
    * Constructs a new {@link Message} posted by {@code user} with {@code text} content and {@code recipient}. Generates a
    * random ID and uses the current system time for the creation time.
    */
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -76,7 +76,7 @@ limitations under the License.
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-language</artifactId>
-      <version>1.55.0</version>
+      <version>1.64.0</version>
```
michaelsafyan, line 79: Thanks for keeping us up-to-date!

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -90,11 +94,29 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String user = userService.getCurrentUser().getEmail();
     String text = Jsoup.clean(request.getParameter("text"), Whitelist.none());
     String recipient = request.getParameter("recipient");
+    System.out.println("before getting sentiment score");
+    System.out.println("text: " + text);
+    float sentimentScore = calculateSentimentScore(text);
```
michaelsafyan, line 99: This isn't really a technical comment so much as a design comment...

It looks like you are putting sentiment analysis in the critical path of posting a message, which may delay the operation of posting a message (and also make it more likely for uploading a message to fail).

Please use the App Engine task queue functionality:
https://cloud.google.com/appengine/docs/standard/java/taskqueue/push/creating-push-queues

Or the more general Cloud PubSub mechanism:
https://cloud.google.com/pubsub/

To schedule sentiment analysis _after_ successfully uploading the message (you may also want to schedule periodic recalculation of sentiment analysis to catch any existing messages and also to keep the sentiment analysis scoring up-to-date with changes/improvements in the sentiment scoring algorithm).

src/main/java/com/google/codeu/data/Message.java
```diff
@@ -26,21 +26,23 @@
...
    * random ID and uses the current system time for the creation time.
    */
-  public Message(String user, String text, String recipient) {
-    this(UUID.randomUUID(), user, text, System.currentTimeMillis(), recipient);
+  public Message(String user, String text, String recipient, float sentimentScore) {
+    this(UUID.randomUUID(), user, text, System.currentTimeMillis(), recipient, sentimentScore);
   }
 
-  public Message(UUID id, String user, String text, long timestamp, String recipient) {
+  public Message(UUID id, String user, String text, long timestamp, String recipient, float sentimentScore) {
```
michaelsafyan, line 39: Since the sentiment score isn't really intrinsic to the Message but is a derived computation about the message, it might be a good idea to create a separate entity and object for it. That is, I would recommend thinking about it/structuring it as a `MessageSentimentScore` references a Message (by key) and contains the computed score (as well as the time at which the sentiment was calculated) and acts like a cache to make it quicker to obtain the score.

For example, if you create another endpoint to obtain the sentiment score of a message, it could act like:

       if exists already computed score for message key:
          return existing score
       compute sentiment score
       store score for future efficiency
       return computed score

There are probably more places where you need just the message content, though, without the score. So it may not make sense to tie the two so tightly.

paulinagzd (approved):

PR #19, opened 2019-03-24
Added image posting capability

shamailah:
Allows users to post image by replacing image urls with img tags. Checks if url is valid before converting it to an img tag. Currently only supports .png and .jpg extensions.

### DIFF
--- a/src/main/java/com/google/codeu/data/Datastore.java
+++ b/src/main/java/com/google/codeu/data/Datastore.java
@@ -66,6 +66,7 @@ public class Datastore {
     messageEntity.setProperty("text", message.getText());
     messageEntity.setProperty("timestamp", message.getTimestamp());
     messageEntity.setProperty("recipient", message.getRecipient());
+    messageEntity.setProperty("sentimentScore", message.getSentimentScore());
 
     datastore.put(messageEntity);
   }
@@ -89,7 +90,7 @@ public class Datastore {
     // Messages between two people, where logged-in user is the recipient
     Filter messagesSentByOtherUser = new Query.FilterPredicate("user", FilterOperator.EQUAL, otherUser);
     Filter messagesReceivedByLoggedInUser = new Query.FilterPredicate("recipient", FilterOperator.EQUAL, loggedInUser);
-    
+
     // All the messages recieved by logged-in user sent by the other user
     Filter otherUserMessages = CompositeFilterOperator.and(messagesSentByOtherUser, messagesReceivedByLoggedInUser);
 
@@ -119,7 +120,9 @@ public class Datastore {
       long timestamp = (long) entity.getProperty("timestamp");
       String sender = (String) entity.getProperty("user");
       String receiver = (String) entity.getProperty("receiver");
-      Message message = new Message(id, sender, text, timestamp, receiver);
+      float sentimentScore = entity.getProperty("sentimentScore") == null ? (float) 0.0 : ((Double) entity.getProperty("sentimentScore")).floatValue();
+
+      Message message = new Message(id, sender, text, timestamp, receiver, sentimentScore);
       messages.add(message);
     }
 
@@ -132,5 +135,4 @@ public class Datastore {
     PreparedQuery results = datastore.prepare(query);
     return results.countEntities(FetchOptions.Builder.withDefaults());
   }
-
 }
--- a/src/main/java/com/google/codeu/servlets/MessageServlet.java
+++ b/src/main/java/com/google/codeu/servlets/MessageServlet.java
@@ -22,6 +22,10 @@ import com.google.appengine.api.users.UserServiceFactory;
 import com.google.codeu.data.Datastore;
 import com.google.codeu.data.Message;
 import com.google.gson.Gson;
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.Document.Type;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.Sentiment;
 import java.io.IOException;
 import java.util.List;
 import javax.servlet.http.HttpServlet;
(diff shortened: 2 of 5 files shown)


paulinagzd (commented):
Handles/validates exceptions well, including the CSS image sizing.

src/main/java/com/google/codeu/data/Datastore.java
```diff
@@ -66,6 +66,7 @@ public void storeMessage(Message message) {
     messageEntity.setProperty("text", message.getText());
     messageEntity.setProperty("timestamp", message.getTimestamp());
     messageEntity.setProperty("recipient", message.getRecipient());
+    messageEntity.setProperty("sentimentScore", message.getSentimentScore());
```
michaelsafyan, line 69: It looks like you have two different changes (image uploading vs sentiment analysis) in the same branch. Can you keep the two features isolated? Make sure when you create new branches that you always start from the master branch.

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -90,11 +97,37 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String user = userService.getCurrentUser().getEmail();
     String text = Jsoup.clean(request.getParameter("text"), Whitelist.none());
     String recipient = request.getParameter("recipient");
+    float sentimentScore = calculateSentimentScore(text);
+
+    // Replace image links with image tags after validating links
+    UrlValidator urlValidator = new UrlValidator();
+    String regex = "(https?://([^\\s.]+.?[^\\s.]*)+/([^\\s.]+.?[^\\s.]*)+.(png|jpg))";
```
michaelsafyan, line 104: It looks like this regex allows both `http` and `https`. You should require strictly `https`, because when the page is `https`, users expected all the data that is being transferred to be encrypted, but the use of `http` resources such as images will result in unencrypted communications happening in the background (e.g. to transfer the image).

michaelsafyan, line 104: It looks like you are relying on the file extension to determine the type. When it comes to URLs, it is much more reliable and robust to use the `Content-Type` response header, instead. For example, an image can just as easily live at:

     http://some.domain.tld/some/path

As at:

     http://some.domain.tld/some/path.jpg

In fact, it is very common (and arguably preferable) for URLs to omit a file extension. And, when it comes to images, it is not uncommon to omit the extension because the actually type of image returned may vary by browser for the same, extension-less URL. A much more robust approach to identifying whether a URL is an image would be to:

Fetch the URL (using the `HEAD` method rather than `GET` method, so that you only fetch the headers and not the entire payload, which you do not need), and then to examine the `Content-Type` header.

The `Content-Type` header will start with `image/` if it is an image (as in `image/jpeg`, `image/png`, etc.)

See also:

- [HTTP HEAD request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD)
- [Content-Type header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)
- [An incomplete list of MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types)

src/main/java/com/google/codeu/servlets/MessageServlet.java
```diff
@@ -90,11 +97,37 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    String replacement = "<img src=\"$1\" />";
```
michaelsafyan, line 105: Rather than inserting this back into the text, you may want to consider creating a separate field of the message for this data. For example, you might want to create a `previewImage` property that is separate from the content.

An advantage to making that property structured is that it then makes it easy to do smarter things with the data without having to process/extract the entire text. For example, it would be possible to identify multiple messages that contain the same image. It would also be possible to prefetch the image and store the image in Cloud Storage along with a reference to the cached image (as additional fields of the `previewImage` property).

src/main/java/com/google/codeu/servlets/MessageServlet.java
michaelsafyan: Since doing this robustly requires a fetch, you may want to consider making this even more expansive than just supporting image URLs, though; for example, articles and other web pages can often have a preview generated from them (like on most social media sites). This is often accomplished with OpenGraph metadata tags like `og:image`.

See also:
  - [The OpenGraph Protocol](http://ogp.me/)

src/main/webapp/css/user-page.css
```diff
@@ -40,8 +40,18 @@
   margin: 5px;
 }
 
+/* Give images in the message body a contained size
+  so they aren't too big */
+.message-body img {
+  display: block;
+  margin-top: 25px;
+  max-width: 250px;
```
michaelsafyan, line 48: I recommend doing this sort of thing with two different elements, an outer `<div>` that represents the "frame" and the inner `` tag that represents the actual image sitting inside the frame, where the `` is vertically and horizontally centered within the frame. You should test that your solution works well for multiple different image aspect ratios. It's often necessary to have different logic for tall vs wide images so that you fix the longer dimension (with `width="100%"` or `height="100%"` on the image tag), and let the shorter-dimension adjust accordingly.

michaelsafyan (commented):
Left a handful of comments. Though I think it would be ideal to act on all or most of this feedback, please feel free to push back if you feel that some of these changes should best be deferred to a separate change (also feel free to leave questions if any of the feedback is unclear).

src/main/java/com/google/codeu/data/Datastore.java
shamailah: Yeah I was confused as to why that was happening because I had created a separate branch. When you say 'start from the master branch' does that mean I should switch to the master branch first and then create a new branch?

michaelsafyan: Yes, that is correct. When you create a new branch, it persists all the context of the branch that you are currently on (so if there are any commits in the current branch, those will also be copied over to the new branch that you create). So, to get a clean branch, one should always start new branches by first switching to the master branch before creating the branch.

shamailah: Ahh I see, I didn't know that. Thanks for clarifying!

paulinagzd (approved):

PR #21, opened 2019-04-19
Styling additional pages

paulinagzd:


markser (commented):
Everything looks good.
If we have time, it would be a good idea to see if we are able to load the images to google app engine so that we can locally build the website instead of having it call to ```https.......``` for the image

markser (approved):

PR #22, opened 2019-04-20
Minimal (might be final) styling touches and corrections

paulinagzd:
…t an extra page with information or articles could be made through HTML


markser (approved):
I am actually going to be merging the stats and the charts page all together so its all under charts.html
I will make sure that the CSS is reflected on the charts.html page. 

Once again, I think it would be worthwhile to upload the image onto something like the blobstore so that if the image goes away from the ```https......``` we still have the copy of the image and wont run into any errors due to image being inaccessible.

